PR #4, opened 2023-07-12
Address trail of bits issues

### LINKED ISSUE #5: TOB Findings 3 & 4
TOB Finding 3:

> DescriptionWhen a refund is registered due to an error, any previous refund balance is overwrittenand lost.Refunds are issued when the bridged Ethereum asset is sent to an incompatible tokenaddress on the Fuel chain.

https://github.com/FuelLabs/fuel-bridge/blob/0861c0c57a3f3125f5ed28e16fdd883624343efd/packages/fungible-token/bridge-fungible-token/src/bridge_fungible_token.sw#L192C14-L192C14

TOB Finding 4:

> DescriptionFunds in the FuelERC20Gateway can be drained due to returning an incorrect asset whenclaiming a refund.

https://github.com/FuelLabs/fuel-bridge/blob/0861c0c57a3f3125f5ed28e16fdd883624343efd/packages/fungible-token/bridge-fungible-token/src/bridge_fungible_token.sw#L132C18-L132C18

DefiCake:
This PR:
- Adds an utility for adding b256 (encoded erc20 amount coming from uint256 of solidity)
- Solves the issues found in trail of bits findings
- Adds new test cases to make sure that we do not experience regression on those findings

Closes #5


packages/fungible-token/bridge-fungible-token/src/bridge_fungible_token.sw
```diff
@@ -190,10 +191,14 @@ impl FRC20 for Contract {
 // Storage-dependant private functions
 #[storage(write)]
 fn register_refund(from: b256, asset: b256, amount: b256) {
-    storage.refund_amounts.get(from).insert(asset, amount);
+    let stored_amount = storage.refund_amounts.get(from).get(asset).try_read().unwrap_or(ZERO_B256);
+
+    // Should not ever overflow if the L1 token is good?
```
K1-R1, line 196: Could you elaborate on what you mean here when saying
> if the L1 token is good?

packages/fungible-token/bridge-fungible-token/tests/harness.rs
```diff
@@ -266,6 +267,137 @@ mod success {
         assert_eq!(amount, config.overflow_2);
     }
 
+    #[tokio::test]
+    async fn claim_refund_of_wrong_token_deposit() {
+        // perform a failing deposit first to register a refund & verify it, then claim and verify output message is created as expected
```
K1-R1, line 272: Can you make this comment a bit clearer?

packages/fungible-token/bridge-fungible-token/tests/utils/environment.rs
```diff
@@ -200,9 +206,10 @@ pub async fn setup_environment(
             &message_sender.into(),
             predicate_root,
             msg.0,
-            message_nonce.into(),
+            Nonce::from(message_nonce),
```
K1-R1, line 209:
```suggestion
            message_nonce,
```

K1-R1 (commented):
Could you run `forc-fmt` and `Cargo clippy --fix`, some of my suggestions will be resolved by these formatters. Overall looks good to me, but I'll await a review from @Braqzen who has worked on this repo most recently

packages/fungible-token/bridge-fungible-token/tests/harness.rs
DefiCake: This is actually not mine - git diffs playing funny here. I can try elaborate the comment but since I 'm not the author I might end up introducing some mistakes

DefiCake: Done

packages/fungible-token/bridge-fungible-token/tests/utils/environment.rs
```diff
@@ -200,9 +206,10 @@ pub async fn setup_environment(
             &message_sender.into(),
             predicate_root,
             msg.0,
-            message_nonce.into(),
+            message_nonce,
             msg.1.clone(),
-        ))
+        ));
+        message_nonce[0] += 1;
```
Braqzen, line 212: Just adding a note because I'm unsure if this is an issue. I believe this will only go up to the `u8` limit so if we ever create (if we even can create that many) more than the limit then this may make the entire suite fail.

Perhaps we should add a comment stating this in case this ever becomes an issue

packages/fungible-token/bridge-fungible-token/tests/harness.rs
```diff
@@ -266,6 +267,141 @@ mod success {
         assert_eq!(amount, config.overflow_2);
     }
 
+    #[tokio::test]
+    async fn claim_refund_of_wrong_token_deposit() {
+        // Send a message informing about a deposit with a a random token address, different from the bridged token
```
Braqzen, line 272:
```suggestion
        // Send a message informing about a deposit with a random token address, different from the bridged token
```

packages/fungible-token/bridge-fungible-token/tests/harness.rs
```diff
@@ -266,6 +267,141 @@ mod success {
...
+        )
+        .await;
+
+        let log_decoder = test_contract.log_decoder();
+        let refund_registered_event = log_decoder
+            .decode_logs_with_type::<RefundRegisteredEvent>(&receipts)
+            .unwrap();
+
+        // Verify the message value was received by the test contract
+        let test_contract_balance = provider
```
Braqzen, line 329: I'm wondering what the 2nd arg here does - the default assetid. Could you remind me? Should we be using some specific asset instead?

packages/fungible-token/bridge-fungible-token/tests/harness.rs
```diff
@@ -266,6 +267,141 @@ mod success {
...
+            &Address::from_str(BRIDGED_TOKEN_GATEWAY).unwrap(),
+            message_receipt.recipient().unwrap()
+        );
+        assert_eq!(message_receipt.amount().unwrap(), 0);
+        assert_eq!(message_receipt.len().unwrap(), 104);
+
+        // message data
+        let (selector, to, token, amount) =
+            env::parse_output_message_data(message_receipt.data().unwrap());
+        assert_eq!(selector, env::decode_hex("0x53ef1461").to_vec());
```
Braqzen, line 398: What hex is this? Should we factor this out into a variable / constant?

packages/fungible-token/bridge-fungible-token/tests/harness.rs
```diff
@@ -1326,4 +1462,130 @@ mod revert {
...
+            .await
+            .unwrap();
+
+        let balance = wallet
+            .get_asset_balance(&AssetId::new(*test_contract_id.hash()))
+            .await
+            .unwrap();
+
+        // Verify the message value was received by the test contract
+        assert_eq!(test_contract_balance, 200);
```
Braqzen, line 1554: I may be misunderstanding. Shouldn't this be `(2 * min_amount) + 1` i.e. 201?
I'm going off the other test being 100

packages/fungible-token/bridge-fungible-token/tests/utils/environment.rs
pixelcircuits: Shouldn't `message_nonce[0]` only overflow at 2^64? `message_nonce` is really just 4 64bit number, right?

Braqzen: I believe it's [u8; 32] under the hood

packages/fungible-token/bridge-fungible-token/src/bridge_fungible_token.sw
```diff
@@ -190,7 +191,11 @@ impl FRC20 for Contract {
 // Storage-dependant private functions
 #[storage(write)]
 fn register_refund(from: b256, asset: b256, amount: b256) {
-    storage.refund_amounts.get(from).insert(asset, amount);
+    let stored_amount = storage.refund_amounts.get(from).get(asset).try_read().unwrap_or(ZERO_B256);
+
+    // Should not ever overflow if the L1 token contract has proper overflow protection mechanisms in place
+    // i.e. we should never get a deposit message that could overflow here if it cannot overflow in the EVM
```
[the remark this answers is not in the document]
pixelcircuits: Let's add the overflow check or at least push it to another issue. Execution is cheap on L2s and this is code we hope never gets hit anyway (only happens if something went wrong when trying to bridge and they need a refund)

packages/fungible-token/bridge-fungible-token/tests/harness.rs
DefiCake: I think it is the native balance. The messages being sent to the L2 contract contain data about the token that is being minted (or burnt) and it is this data which gets relayed to the L1 contract. But these L2 messages can also contain (separated from the data being relayed) a value of the native asset. In this case the each of the test messages carry 100 of the native L2 asset. 

I am unsure about the original intent of this - my best guess right now is that it is being used as some form of check that only the built messages (not more, not less) are being sent to the L2 contract and no other message is changing the state in unexpected ways.

packages/fungible-token/bridge-fungible-token/tests/harness.rs
DefiCake: This is the EVM function selector for `finalizeWithdrawal` on the L1 bridge contract.

```js
/// @notice Finalizes the withdrawal process from the Fuel side gateway contract
    /// @param to Account to send withdrawn tokens to
    /// @param tokenId ID of the token being withdrawn from Fuel
    /// @param amount Amount of tokens to withdraw
    /// @dev Made payable to reduce gas costs
    function finalizeWithdrawal(
        address to,
        address tokenId,
        uint256 amount
    ) external payable whenNotPaused onlyFromPortal {
        require(amount > 0, "Cannot withdraw zero");
        bytes32 fuelTokenId = messageSender();

        //reduce deposit balance and transfer tokens (math will underflow if amount is larger than allowed)
        _deposits[tokenId][fuelTokenId] = _deposits[tokenId][fuelTokenId] - amount;
        IERC20Upgradeable(tokenId).safeTransfer(to, amount);

        //emit event for successful token withdraw
        emit Withdrawal(bytes32(uint256(uint160(to))), tokenId, fuelTokenId, amount);
    }
```

In short, the L2 contract receives this message, which is taken up to the FuelMessagePortal, and executes the data payload, which is a function call to the FuelERC20Gateway contract that holds the funds in L1.

packages/fungible-token/bridge-fungible-token/tests/harness.rs
DefiCake: It sounds to me like it is the native token of the L2. You can see it in the way that messages are built in `environment.rs`: 

```rs
pub async fn construct_msg_data(
    token: &str,
    from: &str,
    to: [u8; 32],
    amount: Unsigned256,
    config: Option<BridgeFungibleTokenContractConfigurables>,
    deposit_to_contract: bool,
    // TODO: https://github.com/FuelLabs/bridge-fungible-token/issues/61
    extra_data: Option<Vec<u8>>,
) -> ((u64, Vec<u8>), (u64, AssetId), Option<ContractId>) {
    let mut message_data = Vec::with_capacity(5);
    message_data.append(&mut decode_hex(token));
    message_data.append(&mut decode_hex(from));
    message_data.append(&mut to.to_vec());
    message_data.append(&mut encode_hex(amount).to_vec());

    let mut deposit_recipient: Option<ContractId> = None;

    if deposit_to_contract {
        let hash = keccak_hash("DEPOSIT_TO_CONTRACT");
        let mut byte: Vec<u8> = vec![0u8];
        byte.copy_from_slice(&hash[..1]);
        message_data.append(&mut byte);
        deposit_recipient = Option::Some(ContractId::new(to));
    };

    if let Some(mut data) = extra_data {
        message_data.append(&mut data);
    };

    let message_data = prefix_contract_id(message_data, config).await;
    let message = (100, message_data);
    let coin = (DEFAULT_COIN_AMOUNT, AssetId::default());

    (message, coin, deposit_recipient)
}
```

There are two fungible amounts in this fn: one is the `amount`, which refers to the bridged asset, and the other is `coin`, which refers to the native asset.

packages/fungible-token/bridge-fungible-token/src/bridge_fungible_token.sw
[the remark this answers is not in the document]
DefiCake: I have added U256 casting which has overflow checks.

packages/fungible-token/bridge-fungible-token/src/bridge_fungible_token.sw
```diff
@@ -33,12 +33,35 @@ use std::{
...
+        asm(output: result, r1: self.a, r2: self.b, r3: self.c, r4: self.d) {
+            sw   output r1 i0; // store the word in r4 in output + 0 words
+            sw   output r2 i1; // store the word in r5 in output + 1 word
+            sw   output r3 i2; // store the word in r5 in output + 1 word
+            sw   output r4 i3; // store the word in r5 in output + 1 word
+        }
+
+        result
+    }
+}
```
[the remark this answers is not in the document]
DefiCake: Added some test in latest commit

packages/fungible-token/bridge-fungible-token/src/cast.sw
```diff
@@ -0,0 +1,82 @@
...
+
+    assert(U256::from(three_b256) == three);
+    assert(U256::from(three_b256) == addition);
+
+    let three_into_b256: b256 = three.into();
+    assert(three_into_b256 == three_b256);
+
+    let addition_into_b256: b256 = addition.into();
+    assert(addition_into_b256 == three_b256);
+}
```
Braqzen, line 48:
```suggestion
fn test_b256_from() {
    // Test the boundary conditions: min, middle, max
    let min = b256::min();
    let middle = 0x000000000000000000000000000000000000000000000000000000000000000a;
    let max = b256::max();
    
    // Alternatively, compare each field of the U256
    assert_eq(U256::from(min), U256::from((0, 0, 0, 0)));
    assert_eq(U256::from(middle), U256::from((0, 0, 0, 10)));
    assert_eq(U256::from(max), U256::from((u64::max(), u64::max(), u64::max(), u64::max())));
}
```

If it complains about an import then add `use core::primitives::*;` at the top.

Test is being re-written to be a unit-test. We are not trying to test the addition of the `U256`. We are testing the conversions of the `from` & `into`

packages/fungible-token/bridge-fungible-token/src/cast.sw
```diff
@@ -0,0 +1,82 @@
...
+    assert(max == U256::from(max_b256)); // test from
+    let random = 0x0123456789abcdef;
+
+    let unsigned = U256::from((random, 0, random, 0));
+    let bits: b256 = 0x0123456789abcdef00000000000000000123456789abcdef0000000000000000;
+
+    let unsigned_into_b256: b256 = unsigned.into();
+    assert(U256::from(bits) == unsigned);
+    assert(unsigned_into_b256 == bits);
+}
```
Braqzen, line 68:
```suggestion
fn test_b256_into() {
    // Test the boundary conditions: min, middle, max
    let min = U256::from((0, 0, 0, 0));
    let middle = U256::from((0, 0, 0, 10));
    let max = U256::from((u64::max(), u64::max(), u64::max(), u64::max()));
    
    assert_eq(min.into(), b256::min());
    assert_eq(middle.into(), 0x000000000000000000000000000000000000000000000000000000000000000a);
    assert_eq(max.into(), b256::max());
}
```

If the compiler cannot infer the type conversion then try assigning to variables and annotating or use the turbofish syntax if it works

packages/fungible-token/bridge-fungible-token/src/cast.sw
```diff
@@ -0,0 +1,82 @@
...
+    let max = U256::from(max_b256);
+    let overflow = max.add(U256::from((0, 0, 0, 1)));
+}
+
+#[test(should_revert)]
+fn test_b256_underflow() {
+    let min_b256: b256 = 0x0000000000000000000000000000000000000000000000000000000000000000;
+    let min = U256::from(min_b256);
+    let overflow = min.subtract(U256::from((0, 0, 0, 1)));
+}
```
Braqzen, line 82:
```suggestion
```

No need, the rest of the standard library should be testing this. The tests above should produce the correct conversion so the rest is covered.

packages/fungible-token/bridge-fungible-token/src/cast.sw
```diff
@@ -0,0 +1,82 @@
+library;
+
+use std::u256::U256;
+use std::constants::ZERO_B256;
+
+impl From<b256> for U256 {
+    fn from(value: b256) -> U256 {
```
Braqzen, line 7:
```suggestion
    fn from(value: b256) -> Self {
```

packages/fungible-token/bridge-fungible-token/src/cast.sw
```diff
@@ -0,0 +1,82 @@
...
+        let (word1, word2, word3, word4) = asm(r1: value) { r1: (u64, u64, u64, u64) };
+        let result = U256::from((word1, word2, word3, word4));
+
+        result
+    }
+
+    fn into(self) -> b256 {
+        let result: b256 = ZERO_B256;
```
Braqzen, line 15:
```suggestion
        let result = b256::min();
```

No need to annotate the type here.
You could use either type afaik.

I'd also remove the import for the constant if you accept these changes which use a different import

packages/fungible-token/bridge-fungible-token/src/cast.sw
DefiCake: Done

Braqzen (approved):

PR #6, opened 2023-07-12
docs: update requirements

luizstacio:
Update requirements on the documents


Braqzen (approved):
Seems fine but I'm wondering why only the docs are changed here and not another file(s) which forces the use of these versions

luizstacio:
> Seems fine but I'm wondering why only the docs are changed here and not another file(s) which forces the use of these versions

I think we should also cover it soon... make sure everything is in sync using the same tooling

K1-R1 (approved):

PR #21, opened 2023-07-19
feat: add auto block generation

luizstacio:
On this PR;
- Add auto block generation on L1 and Fuel Code.
- Change integration tests to wait for the block generation.
- Add deployer config on the `deployAll` script to enable using a different secret on deploy time.
- Use the second privateKey from hardhat to ensure we can use something different from the default signer.


DefiCake (dismissed):

docker/l1-chain/Dockerfile
```diff
@@ -10,7 +10,7 @@ WORKDIR /l1chain
 
 # clone the contracts repo
 ADD ./packages/portal-contracts /l1chain/fuel-v2-contracts
-# copy over the fuel chain and replace consts values
+# copy over the fuel chain replace consts values
```
K1-R1, line 13:
```suggestion
# copy over the fuel chain and replace consts values
```

K1-R1 (approved):
LGTM

K1-R1 (approved):

PR #27, opened 2023-07-27
test: Update Integration Test Proof Set

bvrooman:
Related issues:
- closes https://github.com/FuelLabs/fuel-vm/issues/501

This PR updates the integration tests to use the corrected proof set format for proof sets generated by client Merkle trees. Previously, proof sets contained the hashed leaf data for the leaf undergoing the proof. In order to be compatible with the proof verification algorithm provided by the `fuel-merkle-sol` contracts, we had to remove the leaf data from the proof set. This was done by shifting the proof set (`proofSet.shift()`). Now, the proof sets generated by the client are compatible with the proof verification algorithm by default and this shift is no longer necessary. 

This PR is currently blocked while waiting for an appropriate release of the SDK containing the VM updates.


luizstacio:
@bvrooman I think this fix will be available only when we update everything to fuel-core v0.20, correct?

bvrooman:
> @bvrooman I think this fix will be possible, only when we update everything to fuel-core v0.20, correct?

Yep, that's correct, [fuel-core v0.20.1](https://github.com/FuelLabs/fuel-core/releases/tag/v0.20.1) uses [fuel-vm v0.35.0](https://github.com/FuelLabs/fuel-vm/releases/tag/v0.35.0), and this VM release has the Merkle tree update.

luizstacio:
@bvrooman this is fixed on #34

PR #32, opened 2023-08-02
Documentation of bridge flow

### LINKED ISSUE #15: Organize documentation of architecture of the bridge
- How the flow works?
- What happens if we commit a previous block that is already committed?

DefiCake:
Closes #15


docs/BRIDGE_FLOW.md
```diff
@@ -0,0 +1,304 @@
...
+
+![Block Committing](block_committing.png)
+
+### L1 Deposit
+
+The deposit process is straightforward: the user will send a transaction to deposit an asset in a L1 bridge contract, which will in turn emit a message through the `FuelMessagePortal` to be picked up by the Fuel blockchain 's validator set. These validators will make sure that the message is included in the next blocks of the Fuel blockchain. Message inclusion also carries the possibility of said message being processed by its recipient - in the case of L1 deposit, a message will be delivered from the L1 bridge contract to the L2 bridge contract. This last contract will have the capability of processing these messages, validating that their sender is the L1 bridge contract, and then proceeding to mint an equivalent asset and amount in the Fuel blockchain to the one originally depositted in the L1. Then, the minted assets become available to the user in the L2.
```
Voxelot, line 44: It might be good to clarify that validators simply include messages into the utxo set, and it's up to the user or a third-party service to deliver that message to the appropriate contract. Delivering the message to the contract is enforced by the predicate on the message, so anyone on the chain can submit a tx to process it.

docs/BRIDGE_FLOW.md
```diff
@@ -0,0 +1,304 @@
...
+
+![L1 Deposit](l1_deposit.png)
+
+You can follow the actual implementation of the message processing flow via:
+- [FuelERC20Gateway.sol](../packages/portal-contracts/contracts/messaging/gateway/FuelERC20Gateway.sol) 's `deposit` function:
+```js
+function deposit(bytes32 to, address tokenId, bytes32 fuelTokenId, uint256 amount) external payable whenNotPaused {
```
Voxelot, line 51: These docs may be harder to keep from getting stale if the code ever changes.

docs/BRIDGE_FLOW.md
```diff
@@ -0,0 +1,304 @@
...
+
+To understand this document, it is useful to first get a grasp of a few concepts that will be referred to in the rest of this description.
+
+- Layer 1 (L1): An EVM blockchain, namely Ethereum.
+- Layer 2 (L2): The Fuel blockchain will sometimes be referred as the Layer 2 or L2.
+- Fuel block: a block of the Fuel blockchain, generated by a validator set. It contains transactions, but also messages, outgoing and relayed.
+- Outgoing messages: Messages that are generated in the L1 blockchain that must be sent to L2 entities.
+- Relayed messages: Messages that are generated in the L2 (Fuel) blockchain that must be sent to L1 entities.
+- Fuel epoch: A group of Fuel blocks, packed toghether and identifier by the last block of the epoch.
+- Fuel validators: Entities that validate Fuel blocks and keep the ledger going.
```
Voxelot, line 15: Let's use the term sequencer or sequencer's instead of validators for now

docs/BRIDGE_FLOW.md
```diff
@@ -0,0 +1,304 @@
...
+
+Once the transaction has been included in the Fuel blockchain, the user will need to await two events:
+- First, the Fuel blockchain will need to close an epoch. The epoch will be committed to the L1 blockchain in the [Chain State contract](../packages/portal-contracts/contracts/fuelchain/FuelChainState.sol).
+- Then, once the epoch has been committed, it is needed to wait for its finalization.
+
+After finality has been reached for the epoch that includes the block where the withdrawal transaction was signaled, the user can request the proofs of inclusion of the transaction and message in the block epoch that was comitted to the L1. The user will send a transaction call to the [Message Portal](../packages/portal-contracts/contracts/fuelchain/FuelMessagePortal.sol) 's `relayMessage` function that will check the attached proofs and the finalization status of the epoch that the message belongs to, then proceed to unpack the payload of said message, that should contain an execution instruction to call `finalizeWithdrawal` in  the [L1 Bridge Contract](../packages/portal-contracts/contracts/messaging/gateway/FuelERC20Gateway.sol), releasing the L1 locked tokens.
```
Voxelot, line 193: We could also add additional docs on how these proofs are structured via 2 distinct merkle proofs. One to prove that the block header is included in the history of the epoch, and the other to prove their withdrawal message was included in that proven header.

docs/BRIDGE_FLOW.md
DefiCake: So any specific preference? Off the top of my head, I can think of:

- Removing them completely and only keep the links to the files, maybe mention the functions' ABI, since it is less expected for them to change.
- Keep a reference to a specific code line & commit in the `main` branch.

luizstacio (changes requested):
We should add a link to the summary on the README.md, a new link "concepts" or "architecture" would work.

Braqzen (commented):
We should migrate to use `mdbook` in the future. 
I haven't checked to see if we can import `solidity` code but I assume we could. If that's the case then we can import code snippets back into the docs

Nevertheless, it's a cleaner format and we can segment the pages for each package later.

There's also a part which mentions that the equivalent amount of tokens will be minted on the L2 after a deposit. I always forget this, but we are limited to 9 decimals and we do some bitshifting but we still may lose precision, no?
Should that be mentioned or is that no longer the case?

First time reading through the whole flow and I generally understand most of it

packages/portal-contracts/contracts/fuelchain/FuelChainState.sol
```diff
@@ -89,6 +89,11 @@ contract FuelChainState is Initializable, PausableUpgradeable, AccessControlUpgr
     }
 
     /// @notice Commits a block header.
+    /// @dev Committing to the same commitHeight twice would have the effect of delaying
```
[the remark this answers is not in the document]
DefiCake: I agree. Opened a new issue, resolving this convo for now

Voxelot:
> We should migrate to use `mdbook` in the future. I haven't checked to see if we can import `solidity` code but I assume we could. If that's the case then we can import code snippets back into the docs
> 
> Nevertheless, it's a cleaner format and we can segment the pages for each package later.
> 
> There's also a part which mentions that the equivalent amount of tokens will be minted on the L2 after a deposit. I always forget this, but we are limited to 9 decimals and we do some bitshifting but we still may lose precision, no? Should that be mentioned or is that no longer the case?
> 
> First time reading through the whole flow and I generally understand most of it

This is validated by the sway contract, if the amount doesn't fit into 9 decimals then the message is "refunded" and the user will need to withdraw and try again with the correct amount.

Braqzen:
> > We should migrate to use `mdbook` in the future. I haven't checked to see if we can import `solidity` code but I assume we could. If that's the case then we can import code snippets back into the docs
> > Nevertheless, it's a cleaner format and we can segment the pages for each package later.
> > There's also a part which mentions that the equivalent amount of tokens will be minted on the L2 after a deposit. I always forget this, but we are limited to 9 decimals and we do some bitshifting but we still may lose precision, no? Should that be mentioned or is that no longer the case?
> > First time reading through the whole flow and I generally understand most of it
> 
> This is validated by the [sway contract](https://github.com/FuelLabs/fuel-bridge/blob/main/packages/fungible-token/bridge-fungible-token/src/utils.sw#L54), if the amount doesn't fit into [9 decimals](https://github.com/FuelLabs/fuel-bridge/blob/main/packages/fungible-token/bridge-fungible-token/src/main.sw#L37) then the message is "refunded" and the user will need to withdraw and try again with the correct amount.

Does this mean that we only accept Ethereum contracts that conform to usage of 9 decimals or less, or does the user provide some form of information?

Voxelot:
> Does this mean that we only accept Ethereum contracts that conform to usage of 9 decimals or less, or does the user provide some form of information?

Not necessarily, it still works with standard 18 decimal contracts as long as the precision of the amount transferred isn't higher than 9 decimals. The bridging dapp/user should truncate the withdrawal amount to 9 decimals before initiating a deposit.

Braqzen:
> > Does this mean that we only accept Ethereum contracts that conform to usage of 9 decimals or less, or does the user provide some form of information?
> 
> Not necessarily, it still works with standard 18 decimal contracts as long as the precision of the amount transferred isn't higher than 9 decimals. The bridging dapp/user should truncate the withdrawal amount to 9 decimals before initiating a deposit.

So the following would bridge fine: `1.0000000010` (10 d.p. but last is 0)
but the following would lose value: `1.0000000011` (10 d.p. but last is 1).

I see, that is worrying as that means the user is responsible for a technical limitation.
Typically, they would click `max` on some UI component which may have a greater precision and thus they would lose some dust

DefiCake:
Updated the PR with the requested changes and a small intro to the decimals issue.

docs/ARCHITECTURE.md
```diff
@@ -88,4 +88,10 @@ You can follow the implementation of this flow via:
...
+
+Additionally, it is a de-facto practice in the space to use 18 decimals to represent the token, whereas in Fuel, 9 decimals are used to represent amounts under the unit. This means that there is a potential loss of precision when it comes to translating L1 amounts to L2 amounts, as the user could lose some dust amounts between bridging operations. Ideally, this limitation could be worked around as long as the deposits initiated in L1 do not specify amounts under 9 decimals (i.e. 1.0000000010 does not lose amounts, 1.0000000011 would lose 0.0000000001). DApps that enable bridge operations should observe this limitation and truncate the amounts to avoid loss of precision.
```
Braqzen, line 97:
```suggestion
Additionally, it is a de-facto practice in the space to use 18 decimals to represent the token, whereas in Fuel, 9 decimals are used to represent amounts under the unit. This means that there is a potential loss of precision when it comes to translating L1 amounts to L2 amounts, as the user could lose some dust amounts between bridging operations. Ideally, this limitation could be worked around as long as the deposits initiated in L1 do not specify amounts greater than 9 decimals (i.e. 1.0000000010 does not lose amounts, 1.0000000011 would lose 0.0000000001). DApps that enable bridge operations should observe this limitation and truncate the amounts to avoid loss of precision.
```

I think this is meant to be greater, correct?

docs/ARCHITECTURE.md
```diff
@@ -88,4 +88,10 @@ You can follow the implementation of this flow via:
...
 - [Message Portal](../packages/portal-contracts/contracts/fuelchain/FuelMessagePortal.sol) 's `relayMessage` function
-- [FuelERC20Gateway.sol](../packages/portal-contracts/contracts/messaging/gateway/FuelERC20Gateway.sol) 's `finalizeWithdrawal`
\ No newline at end of file
+- [FuelERC20Gateway.sol](../packages/portal-contracts/contracts/messaging/gateway/FuelERC20Gateway.sol) 's `finalizeWithdrawal`
+
+## Decimals adjustment
+
+Easing the operation and participation of light clients in the network without compromising on security is a key principle in Fuel. In that sense, Fuel strives to minimize the size of the information that is exchanged between peers through the protocol. One of the design decisions furthering that goal is the use of 64 bits (`u64`) to code the balances that are bridged from L1 to L2. The most popular token standard, `ERC20`, codes the amounts with `256 bits`. If the amount transferred from L1 to L2 surpasses the capacity of Fuel to mint the L2 counterpart, a `refund` message will be generated, that will allow the user to retrieve the originally depositted amount in L1.
```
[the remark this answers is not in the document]
DefiCake: All 10 units would be refunded

Braqzen (approved):
In the future we could reword some of the content to be concise and digestible in smaller chunks but it's probably fine for now

luizstacio (approved):

PR #34, opened 2023-08-15
feat: update compatibility with fuel-core v0.20

luizstacio:
close: #17


docker/block-committer/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM ghcr.io/fuellabs/fuel-block-committer
+FROM ghcr.io/fuellabs/fuel-block-committer:sha-f76f96f
```
Voxelot, line 1:
```suggestion
FROM ghcr.io/fuellabs/fuel-block-committer:v0.1.0
```

packages/fungible-token/bridge-fungible-token/fuel-toolchain.toml
```diff
@@ -2,5 +2,5 @@
 channel = "latest-2023-07-20"
 
 [components]
-forc = "0.42.1"
-fuel-core = "0.19.1"
+forc = "0.44.0"
+fuel-core = "0.20.3"
```
Voxelot, line 6:
```suggestion
fuel-core = "0.20.4"
```

packages/fungible-token/bridge-fungible-token/src/data_structures.sw
```diff
@@ -15,7 +15,7 @@ impl MessageData {
     ///
     /// any data beyond 160 bytes means deposit is meant for a contract.
     /// if data is > 161 bytes, then we also need to call process_message on the destination contract.
-    pub fn parse(msg_idx: u8) -> Self {
+    pub fn parse(msg_idx: u64) -> Self {
```
Voxelot, line 18: why did we change this to u64?

packages/fungible-token/bridge-fungible-token/fuel-toolchain.toml
luizstacio: If I change it to 0.20.4 I got an error `component 'fuel-core' not found in currently active toolchain 'latest-2023-07-20-aarch64-apple-darwin'`. This is probably because the 0.20.4 is in another channel but I didn't find one that works.

Voxelot: ok no worries

packages/fungible-token/bridge-fungible-token/src/data_structures.sw
Voxelot: ah this is likely because sway doesn't allow for implicit integer casting now? Since I don't think the stdlib changed here. The strange thing to me is that it seems like we should be able to safely upcast ints without any concerns for precision.

cc @IGI-111

packages/fungible-token/bridge-fungible-token/src/main.sw
```diff
@@ -74,8 +75,10 @@ impl MessageReceiver for Contract {
                 register_refund(message_data.from, message_data.token, message_data.amount);
             },
             Result::Ok(amount) => {
+                let subId: SubId = ZERO_B256;
+                let tokenId = sha256((contract_id(), subId));
```
Voxelot, line 79: [nit] there is a helper function for this in the stdlib now  `std::asset_id::construct_default_asset_id`.

Voxelot (approved):

packages/fungible-token/bridge-fungible-token/src/data_structures.sw
IGI-111: Yeah we have safe upcasts, though they have to be explicitly called.

I'm not sure if this should be fixed here or on the library side however.

packages/fungible-token/bridge-fungible-token/Cargo.toml
```diff
@@ -6,8 +6,8 @@ edition = "2021"
 license = "Apache-2.0"
 
 [dependencies]
-fuel-core-types = { version = "0.19.1", default-features = false }
-fuels = { version = "0.44", features = ["fuel-core-lib"] }
+fuel-core-types = { version = "0.20.4", default-features = false }
+fuels = { version = "0.45.1", features = ["fuel-core-lib"] }
```
Braqzen, line 10: Version 46 is out

packages/fungible-token/bridge-fungible-token/tests/functions/bridge/mod.rs
```diff
@@ -63,14 +64,16 @@ mod success {
         )
         .await;
 
+        println!("receipts: {:#?}", receipts);
```
Braqzen, line 67: We ought to keep debugging statements out of tests

packages/fungible-token/bridge-fungible-token/tests/functions/bridge/mod.rs
```diff
@@ -115,7 +118,7 @@ mod success {
             message_receipt.recipient().unwrap()
         );
         assert_eq!(message_receipt.amount().unwrap(), 0);
-        assert_eq!(message_receipt.len().unwrap(), 104);
+        assert_eq!(message_receipt.len().unwrap(), 100);
```
Braqzen, line 121: What changed about the len of the receipts?

Voxelot: This is likely due to a correction in the serialized format of message receipts: https://github.com/FuelLabs/fuel-vm/commit/18a2e4a13094448068e378baf3b89e448ec356d6

Voxelot: Ah it was also changed here: https://github.com/FuelLabs/fuel-bridge/pull/34/files#diff-73b34b7aa5b07f5607f9ebcf819a8d15226bb84d5553ca531373dbeb9104cabeR8

PR #39, opened 2023-08-18
feat: add utility to grant comitter role to other addresses

DefiCake:


packages/portal-contracts/scripts/grantComitterRole.ts
```diff
@@ -0,0 +1,43 @@
...
+    const provider = new StaticJsonRpcProvider(RPC_URL);
+
+    const admin = new ethers.Wallet(DEPLOYER_KEY, provider);
+
+    const target = FuelChainState__factory.connect(FUEL_CHAIN_STATE_ADDRESS, admin);
+    
+    const role = await target.callStatic.COMMITTER_ROLE();
+
+    await target.grantRole(role, COMITTER_ADDRESS).then(tx => {
+        console.log("\t> Transaction sent with hash", tx.hash, "... waiting confirmation");
```
Braqzen, line 35:
```suggestion
        console.log("\t> Transaction sent with hash", tx.hash, "... awaiting confirmation");
```

Wondering if we ought to use logs or remove them. Probably doesn't matter

Braqzen (commented):
To grant to another address we must set it up as an environment variable.
Meaning, we have to keep changing it per new address.

Since this is an environment variable instead of a parameter it implies that we would rarely use this. Is that the case?
Should this be expanded to take any number of addresses via params?

DefiCake:
@Braqzen I agree that it would better fit a pattern with params instead of `envs`. Unfortunately hardhat scripts do need `envs`, https://hardhat.org/hardhat-runner/docs/advanced/scripts , https://stackoverflow.com/questions/69111785/hardhat-run-with-parameters . 

We could switch to `task` based scripts, which allow passing parameters, but this requires a little bit more of boilerplate and I was on the run with this. I 'd suggest we merge this and open an issue to refactor the current scripting to tasks.

Braqzen (dismissed):

packages/portal-contracts/scripts/grantComitterRole.ts
```diff
@@ -0,0 +1,43 @@
+import { Signer, ethers } from "ethers";
+import { isAddress } from "ethers/lib/utils";
+import {FuelChainState__factory} from '../typechain';
+import { StaticJsonRpcProvider } from "@ethersproject/providers";
+
+const main = async () => {
+    const {DEPLOYER_KEY, FUEL_CHAIN_STATE_ADDRESS, COMITTER_ADDRESS, RPC_URL} = process.env;
+
+    if(!DEPLOYER_KEY) {
+        throw new Error('No DEPLOYER_KER env var');
```
luizstacio, line 10:
```suggestion
        throw new Error('No DEPLOYER_KEY env var');
```

luizstacio (commented):
I think we should add a script to the package.json

luizstacio (approved):

luizstacio:
@DefiCake I think just need to run a `pnpm format`

luizstacio (approved):

PR #40, opened 2023-08-21
Add NFT Support

### LINKED ISSUE #31: NFT bridging support

DefiCake:
This PR implements NFT support and general `token_id` / `asset_sub_id` awareness for the L2 component of the bridge.

Closes #31


packages/portal-contracts/contracts/messaging/gateway/FuelERC20Gateway.sol
```diff
@@ -92,109 +94,120 @@ contract FuelERC20Gateway is
...
-            bytes32(uint256(uint160(tokenId))),
-            bytes32(uint256(uint160(msg.sender))), //from
-            to,
-            bytes32(amount)
+            fuelContractId,
+            bytes32(uint256(uint160(tokenAddress))), // OFFSET_TOKEN_ADDRESS = 32
+            bytes32(0), // OFFSET_TOKEN_ID = 64
+            bytes32(uint256(uint160(msg.sender))), //from, OFFSET_FROM = 96
+            to, // OFFSET_TO = 128
+            amount // OFFSET_AMOUNT = 160
```
Braqzen, line 121: This is no longer wrapped in `bytes`, is there a reason?

packages/portal-contracts/contracts/messaging/gateway/FuelERC20Gateway.sol
```diff
@@ -92,109 +94,120 @@ contract FuelERC20Gateway is
...
     /// @dev Made payable to reduce gas costs
     function depositWithData(
         bytes32 to,
-        address tokenId,
-        bytes32 fuelTokenId,
+        address tokenAddress,
+        bytes32 fuelContractId,
         uint256 amount,
-        bytes memory data
+        bytes calldata data
```
Braqzen, line 138: aaaaaaah, gas optimizationsss :)

Braqzen (commented):
I have glanced through the linked issue and it mentions the strong types for contracts, assets and subids.

I'm leaning towards using the strict types.
I haven't thought about the NFT.

packages/portal-contracts/contracts/messaging/gateway/FuelERC20Gateway.sol
DefiCake: `abi.encodePacked` just concatenates the values in its bytes representation. An uint256 is equivalent to bytes32 in that regard, so wrapping it is redundant.

packages/fungible-token/bridge-fungible-token/src/data_structures.sw
```diff
@@ -7,32 +7,44 @@ pub struct MessageData {
     from: b256,
     len: u16,
     to: Identity,
-    token: b256,
+    token_address: b256,
+    token_id: b256,
 }
 
+const OFFSET_TOKEN_ADDRESS = 32;
```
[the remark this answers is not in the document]
DefiCake: Done 👍🏻

packages/portal-contracts/contracts/messaging/gateway/FuelERC721Gateway.sol
```diff
@@ -0,0 +1,202 @@
...
+    PausableUpgradeable,
+    AccessControlUpgradeable,
+    UUPSUpgradeable
+{
+    ////////////
+    // Events //
+    ////////////
+
+    /// @dev Emitted when tokens are deposited from Ethereum to Fuel
+    event Deposit(bytes32 indexed sender, address indexed tokenAddress, bytes32 fuelContractId, uint256 tokenId);
```
Braqzen, line 31: Wondering if we want to easily index via the contract id too?
Similarly, in withdrawal below.

packages/portal-contracts/contracts/messaging/gateway/FuelERC721Gateway.sol
```diff
@@ -0,0 +1,202 @@
...
+    function finalizeWithdrawal(
+        address to,
+        address tokenAddress,
+        uint256 /*amount*/,
+        uint256 tokenId
+    ) external payable override whenNotPaused onlyFromPortal {
+        bytes32 fuelContractId = messageSender();
+        require(_deposits[tokenAddress][tokenId] == fuelContractId, "Fuel bridge does not own this token");
+
+        delete _deposits[tokenAddress][tokenId];
```
Braqzen, line 167: Is it cheaper to `delete` or override with a default value? Is it safe to use some sentinel value as a default?

Braqzen (commented):
Have not looked over `erc721Gateway.test.ts`. Will try to get around to it by Thursday

packages/fungible-token/bridge-fungible-token/src/events.sw
```diff
@@ -19,3 +20,10 @@ pub struct WithdrawalEvent {
     from: Identity,
     to: b256,
 }
+
+pub struct ClaimRefundEvent {
+    amount: b256,
+    originator: b256,
```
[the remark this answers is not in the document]
DefiCake: It isn't, naming was cloned from the argument of the function that generates it. I think I 'll change both to `from`

packages/portal-contracts/test/erc20Gateway.ts
```diff
@@ -822,7 +861,8 @@ describe('ERC20 Gateway', async () => {
           blockInRoot,
           msgInBlock
         )
-      ).to.be.revertedWith('Message relay failed');
+      ).to.be.reverted;
```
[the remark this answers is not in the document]
DefiCake: I have changed the way the FuelMessagePortal relays the messages so that it bubbles up the reason instead of giving a generic message. Also removed the use of the library it was using before because it was giving us no benefits over a normal call.

Some of these tests now fail with an underflow / overflow at the compiler level and I have not found a way to specify those with the current set of hardhat chai matchers. I 'll give it another look

packages/integration-tests/tests/bridge_erc721.ts
```diff
@@ -0,0 +1,290 @@
...
+        env.fuel.provider,
+        fuelTokenMessageReceiver,
+        fuelTokenMessageNonce,
+        FUEL_MESSAGE_TIMEOUT_MS
+      );
+      expect(message).to.not.be.null;
+      const tx = await relayCommonMessage(env.fuel.deployer, message);
+      const result = await tx.waitForResult();
+
+      expect(result.status).to.equal('success');
```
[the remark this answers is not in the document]
DefiCake: Inherited from the erc20 tests. In the next `it` block, we check that the message actually succeeded the way we wanted:

```js
it('Check ERC721 arrived on Fuel', async () => {
      // check that the recipient balance has increased by the expected amount
      const balance = await fuelTokenReceiver.getBalance(fuel_testAssetId);
      expect(balance.toNumber()).to.be.eq(1);
    });
```

We could also add a check for the `log DepositEvent {...}`

packages/integration-tests/tests/bridge_erc721.ts
```diff
@@ -0,0 +1,290 @@
...
+      );
+      withdrawMessageProof = await fuelTokenSender.provider.getMessageProof(
+        tx.id,
+        messageOutReceipt.messageId,
+        commitHashAtL1
+      );
+
+      // check that the sender balance has decreased by the expected amount
+      let newSenderBalance = await fuelTokenSender.getBalance(fuel_testAssetId);
+      expect(newSenderBalance.toNumber()).to.be.eq(0);
```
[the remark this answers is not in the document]
DefiCake: It 'd be redundant, the precondition is already checked:

```js
it('Check ERC721 arrived on Fuel', async () => {
      // check that the recipient balance has increased by the expected amount
      const balance = await fuelTokenReceiver.getBalance(fuel_testAssetId);
      expect(balance.toNumber()).to.be.eq(1);
    });
```

packages/portal-contracts/contracts/messaging/gateway/FuelERC721Gateway.sol
DefiCake: OK for me - I 'd do it in the ERC20 gateway too for the sake of consistency

packages/portal-contracts/contracts/messaging/gateway/FuelERC721Gateway.sol
DefiCake: Depends on the context:
- Deleting (or setting from non-0 to 0 value, it is the same action) has the effect of releasing storage space for the blockchain, which is rewarded with decreased gas costs - ie. the specific withdrawal transaction would be cheaper. WRT the use of `delete` instead of assigning a null value, it is purely opinionated - I tend to use `delete` because I feel it is more idiomatic.
-  Setting a non-0 value to another non-0 (sentinel) value is cheaper than going from 0 to non-0. This would make withdrawal a bit more expensive, and deposit a bit cheaper _when the asset has already been bridged before_. Also we 'd need to check for both the null value and the sentinel value, because we cannot initialise storage to a default / sentinel value.

Overall I think the sentinel value would be a bit more expensive long term.

packages/fungible-token/bridge-fungible-token/src/interface.sw
```diff
@@ -52,4 +52,8 @@ abi FungibleBridge {
 
     /// Get the address of the gateway that holds the bridged tokens
     fn bridged_token_gateway() -> b256;
+
+    // Gets the sub_id component of the preimage of an asset_id
```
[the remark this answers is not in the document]
DefiCake: Reworded it

packages/fungible-token/bridge-fungible-token/src/interface.sw
```diff
@@ -27,7 +27,7 @@ abi FungibleBridge {
     /// * `originator`: [b256] - the address entitled to a refund
     /// * `asset`: [b256] - the token to be refunded back to the originator
     #[storage(read, write)]
-    fn claim_refund(originator: b256, asset: b256);
+    fn claim_refund(originator: b256, token_address: b256, token_id: b256);
```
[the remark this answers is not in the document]
DefiCake: Done

packages/portal-contracts/contracts/messaging/gateway/FuelERC721Gateway.sol
DefiCake: Done

packages/fungible-token/bridge-fungible-token/src/interface.sw
```diff
@@ -24,10 +24,11 @@ abi FungibleBridge {
     ///
     /// # Arguments
     ///
-    /// * `originator`: [b256] - the address entitled to a refund
-    /// * `asset`: [b256] - the token to be refunded back to the originator
+    /// * `from`: [b256] - the depositor 's address entitled to a refund
```
K1-R1, line 27:
```suggestion
    /// * `from`: [b256] - the depositor's address entitled to a refund
```

packages/fungible-token/bridge-fungible-token/src/main.sw
```diff
@@ -117,36 +125,46 @@ impl MessageReceiver for Contract {
 
 impl FungibleBridge for Contract {
     #[storage(read, write)]
-    fn claim_refund(originator: b256, asset: b256) {
-        let stored_amount = storage.refund_amounts.get(originator).get(asset).read();
-        require(stored_amount != ZERO_B256, BridgeFungibleTokenError::NoRefundAvailable);
+    fn claim_refund(from: b256, token_address: b256, token_id: b256) {
+        let asset = sha256((token_address, token_id));
+        let amount = storage.refund_amounts.get(from).get(asset).read();
```
K1-R1, line 130:
```suggestion
        let amount = storage.refund_amounts.get(from).get(asset).try_read().unwrap_or(ZERO_B256);
```
If no value has been inserted to the key `asset`, then reading will return `None`. The suggestion is only necessary when we're trying to read from a key that may not yet have had a value inserted, in which case we should always use `.try_read()`. 
This applies to other instances in this file too, where `read` should be replaced with `try_read` when using `get` on `StorageMap`

K1-R1 (commented):
Not a part of this PR; but is FRC20  in [the interface file](https://github.com/FuelLabs/fuel-bridge/blob/450a0dbae021ce4a78fdb2fa3b73e1c129b46e14/packages/fungible-token/bridge-fungible-token/src/interface.sw) supposed to follow the [SRC-20 Token Standard](https://github.com/FuelLabs/sway-standards/tree/master/standards/src_20)? If so we'll need to update this interface to follow the standard.

packages/fungible-token/bridge-fungible-token/src/main.sw
DefiCake: Done

packages/portal-contracts/test/erc721Gateway.test.ts
```diff
@@ -0,0 +1,1044 @@
...
+
+        expect(
+          await env.fuelERC721Gateway.tokensDeposited(env.nft.address, tokenId)
+        ).to.be.equal(fuelTarget);
+      };
+
+      // Deposit to fuelTokenTarget1
+      {
+        const tokenId = 0;
+        await testDeposit(
```
Braqzen, line 539: Should these have assertions?

packages/fungible-token/bridge-fungible-token/src/main.sw
```diff
@@ -188,15 +208,29 @@ impl FRC20 for Contract {
...
         amount,
     });
 }
+
+#[storage(read)]
+fn _asset_to_sub_id(asset_id: b256) -> b256 {
+    let sub_id = storage.asset_to_sub_id.get(asset_id).try_read();
+    require(sub_id.is_some(), BridgeFungibleTokenError::AssetNotFound);
+    sub_id.unwrap()
+}
```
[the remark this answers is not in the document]
DefiCake: Ran forc fmt, it added the new lines here and in other files

packages/fungible-token/bridge-fungible-token/tests/functions/bridge/mod.rs
```diff
@@ -522,4 +556,131 @@ mod revert {
...
+
+        let asset_balance =
+            contract_balance(provider, bridge.contract_id(), AssetId::default()).await;
+        let balance = wallet_balance(&wallet, &get_asset_id(bridge.contract_id())).await;
+
+        // Verify the message value was received by the bridge contract
+        assert_eq!(asset_balance, MESSAGE_AMOUNT);
+
+        // Verify that no tokens were minted for message.data.to
+        assert_eq!(balance, 0);
```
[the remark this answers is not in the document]
DefiCake: Done 👍🏻

packages/portal-contracts/test/erc721Gateway.test.ts
DefiCake: It 'd be redundant. `ethers` throws when the transaction is not successful

packages/integration-tests/tests/bridge_erc721.ts
```diff
@@ -0,0 +1,290 @@
...
+        fromBlock: result.blockNumber,
+        toBlock: result.blockNumber,
+      });
+
+      expect(rest.length).to.be.equal(0);
+
+      // parse events from logs
+      let event = env.eth.fuelMessagePortal.interface.parseLog(log);
+      fuelTokenMessageNonce = new BN(event.args.nonce.toHexString());
+      fuelTokenMessageReceiver = Address.fromB256(event.args.recipient);
```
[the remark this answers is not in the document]
luizstacio: I think we need some code improvements, when doing the code, I have also followed the previous standards. But is something common to have global vars on the tests.

packages/integration-tests/tests/bridge_erc721.ts
```diff
@@ -0,0 +1,290 @@
...
+        null
+      );
+
+      const [log, ...rest] = await env.eth.provider.getLogs({
+        ...filter,
+        fromBlock: result.blockNumber,
+        toBlock: result.blockNumber,
+      });
+
+      expect(rest.length).to.be.equal(0);
```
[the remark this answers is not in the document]
DefiCake: I think I overlooked this comment, but `rest` is the rest of elements in the array returned by `getLogs`. In this case we are checking that no other logs were emitted. Inherited from the pattern of tests of erc20.

packages/fungible-token/bridge-fungible-token/src/main.sw
```diff
@@ -61,9 +62,17 @@ impl MessageReceiver for Contract {
         let message_data = MessageData::parse(msg_idx);
         require(message_data.amount != ZERO_B256, BridgeFungibleTokenError::NoCoinsSent);
 
+        let sub_id: SubId = message_data.token_id;
```
Braqzen, line 65:
```suggestion
        let sub_id = message_data.token_id;
```

Explicit typing here is unnecessary

packages/fungible-token/bridge-fungible-token/src/data_structures.sw
```diff
@@ -2,12 +2,21 @@ library;
...
+pub const CONTRACT_DEPOSIT_WITHOUT_DATA_LEN = 193u16;
+
 pub struct MessageData {
     amount: b256,
     from: b256,
     len: u16,
     to: Identity,
-    token: b256,
+    token_address: b256,
+    token_id: b256,
```
luizstacio, line 19:
```suggestion
    asset_id: b256,
    asset_sub_id: b256,
```
Maybe this align better with how we call things on the chain

packages/integration-tests/tests/bridge_erc721.ts
```diff
@@ -0,0 +1,290 @@
...
+      env,
+      eth_testToken,
+      env.eth.fuelERC721Gateway,
+      FUEL_TX_PARAMS,
+      0
+    );
+    fuel_testContractId = fuel_testToken.id.toHexString();
+  });
+
+  it('Setup tokens to bridge', async () => {
```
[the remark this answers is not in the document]
luizstacio: Yeah all setup stages normally stay on before all action

luizstacio (dismissed):

packages/fungible-token/bridge-fungible-token/tests/functions/bridge/mod.rs
```diff
@@ -458,6 +489,8 @@ mod success {
 mod revert {
     use super::*;
     use crate::utils::setup::get_asset_id;
+    use crate::utils::setup::RefundRegisteredEvent;
```
Braqzen, line 492: Since these both come from the same place (setup) I would import them on 1 line with `{get_asset_id, RefundRegisteredEvent}`

packages/solidity-contracts/contracts/messaging/gateway/FuelERC721Gateway.sol
```diff
@@ -0,0 +1,212 @@
...
+        address tokenAddress,
+        uint256 /*amount*/,
+        uint256 tokenId
+    ) external payable override whenNotPaused onlyFromPortal {
+        bytes32 fuelContractId = messageSender();
+        require(_deposits[tokenAddress][tokenId] == fuelContractId, "Fuel bridge does not own this token");
+
+        delete _deposits[tokenAddress][tokenId];
+
+        IERC721Upgradeable(tokenAddress).transferFrom(address(this), to, tokenId);
```
Braqzen, line 179: Instantiating this instance each time to transfer is costly. Should we create a mapping of instances, fetch it from storage and then transfer?
That may or may not be cheaper over time.

packages/solidity-contracts/test/erc721Gateway.test.ts
```diff
@@ -0,0 +1,1036 @@
...
+  generateBlockHeaderLite,
+} from '../protocol/blockHeader';
+import { EMPTY, ZERO } from '../protocol/constants';
+import { setupFuel } from '../protocol/harness';
+import type { HarnessObject } from '../protocol/harness';
+import Message, { computeMessageId } from '../protocol/message';
+import { randomAddress, randomBytes32, tai64Time } from '../protocol/utils';
+import type { NFT } from '../typechain';
+
+const CONTRACT_MESSAGE_PREDICATE =
```
Braqzen, line 22: Should we dynamically load this instead?

Braqzen (commented):
What is the purpose of the following `.md` files?

- `heavy-cobras-hammer`
- `smooth-trees-explain`

Just some comments / nits

packages/solidity-contracts/contracts/messaging/gateway/FuelERC721Gateway.sol
DefiCake: This is not an instantiation / creation. It is a call to another contract. Retrieving it from a mapping would be costlier. This has no gas cost to my knowledge.

In assembly code this would translate to `call(tokenAddress, some ugly params, area of memory where to store result, size of memory of result...)`

packages/fungible-token/bridge-fungible-token/tests/functions/bridge/mod.rs
DefiCake: actually deleting it since it is not used

DefiCake:
> What is the purpose of the following `.md` files?
> 
> * `heavy-cobras-hammer`
> * `smooth-trees-explain`
> 
> Just some comments / nits

They are artifacts generated by the versioning tool added by @luizstacio . They are automatically deleted afterwards with a PR launched from a bot account (I think). Luiz might have better knowledge than I do.

packages/integration-tests/tests/bridge_erc721.ts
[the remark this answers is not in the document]
DefiCake: Done

luizstacio:
> * `heavy-cobras-hammer`
> * `smooth-trees-explain`

Yeah they are used to generate the change logs.

luizstacio (approved):

Braqzen (approved):

PR #67, opened 2023-09-16
feat: configure packages build to release

luizstacio:
On this PR;
- Rename package `@fuel-bridge/portal-contracts` to `@fuel-bridge/solidity-contracts` as we release this on npm it looks correct to add a name that matches better with the nature of the contracts as they are not fuel native but instead focused on solidity. And if in the future we add contracts in other langs they will not need to be renamed.
- Implement configurations for build and release packages `@fuel-bridge/solidity-contracts`, `@fuel-bridge/message-predicates` and `@fuel-bridge/test-utils`.
- Create a new package `esbuild-bin-loader` that implements a simple esbuild plugin to convert `.bin` into js modules.

Next Step;
- IImplement changeset release tool to versioning and add the CI to release the packages.


DefiCake (approved):
lgtm, but I have one thing to ask - the context being that hardhat for example does not still have support for ESM + typescript at the same time (you can use ESM in JS hardhat projects)

Will this package have some sort of support for commonjs? Or are we restricting it to ESM projects?

luizstacio:
> lgtm, but I have one thing to ask - the context being that hardhat for example does not still have support for ESM + typescript at the same time (you can use ESM in JS hardhat projects)
> 
> Will this package have some sort of support for commonjs? Or are we restricting it to ESM projects?

On the contracts package, I'm exporting the artifacts, contracts, and the ABIs only, this means I'm thinking of exporting any hardhat configs at this time. But all the exported things have support to modules and commonjs.

PR #80, opened 2023-09-30
ERC20 unit test refactor

DefiCake:
This PR:
- Removes redundancy around the erc20 tests
- upgrades hardhat (increased setup speed)
- Moves some code around for better reusability
- Discards a part of `getOrDeployERC721Gateway` that was pasted from ERC20 gateway by mistake


packages/solidity-contracts/protocol/utils.ts
```diff
@@ -1,6 +1,9 @@
 // Helper functions for testing
-import { BigNumber as BN } from 'ethers';
+import type { BigNumberish, BytesLike } from 'ethers';
+import { BigNumber as BN, BigNumber, ethers } from 'ethers';
```
Braqzen, line 3:
```suggestion
import type { BigNumberish, BytesLike } from 'ethers';
import { BigNumber as BN, ethers } from 'ethers';
```

Haven't seen `import type` before. Can these two imports be merged?

packages/solidity-contracts/contracts/test/MockFuelMessagePortal.sol
```diff
@@ -0,0 +1,17 @@
...
+
+    function setMessageSender(bytes32 value) external {
+        _messageSender = value;
+    }
+
+    function messageSender() external view returns (bytes32) {
+        return _messageSender;
+    }
+
+    function sendMessage(bytes32, bytes calldata) external pure {}
```
Braqzen, line 16: What purpose does this empty function serve?

packages/solidity-contracts/protocol/utils.ts
```diff
@@ -49,3 +52,90 @@ export function tai64Time(millis: number): string {
...
+        [
+          fuelContractId,
+          tokenAddress,
+          BigNumber.from(tokenId),
+          from,
+          to,
+          amount,
+          depositToContractFlag,
+        ]
+      );
```
Braqzen, line 90: For code quality (CQ) I would assign these two arrays to variables at the top. In the `if` expression I would check if the length is not 0. If the length is not 0 then I would append the `bytes` and `data` to the arrays.

At the end I would throw them into a singular return.

```typescript
let types = ["bytes32", ...];
let messageData = [fuelContractId, ...];

if (data.length > 0) {
    // append to arrays
}

return ethers.utils.solidityPack(types, messageData);
```

I see there's an extra return for when the data is not provided so the variables could be factored out to the top.

packages/solidity-contracts/test/erc20Gateway.ts
```diff
@@ -1,1009 +1,413 @@
...
 
   describe('Make both valid and invalid ERC20 deposits', async () => {
-    let provider: Provider;
-    before(async () => {
-      provider = env.fuelMessagePortal.provider;
-    });
+    // let provider: Provider;
+    // before(async () => {
+    //   provider = env.fuelMessagePortal.provider;
+    // });
```
Braqzen, line 85: Should this be removed?

packages/solidity-contracts/test/erc20Gateway.ts
Braqzen, line 1: It seems like some of the deposit tests previously checked the before and then after conditions while now the tests state the exact expected value instead. This means that the value could have remained the same because of some bug, consistently across all tests. I would add back in the checks for the balance to ensure it has changed.

Some role based tests have also been removed. Is there a reason?

DefiCake: > It seems like some of the deposit tests previously checked the before and then after conditions while now the tests state the exact expected value instead. This means that the value could have remained the same because of some bug, consistently across all tests. I would add back in the checks for the balance to ensure it has changed.

I have reviewed the tests to check if I missed a case, but I did not see any. All the "delta" balance checks have been replaced by the hardhat helper `expect(...).to. changeTokenBalance(...)`. In revert cases, they have been removed, the reasoning being that it does not make sense to check state changes if a transaction has been completely reverted.

> Some role based tests have also been removed. Is there a reason?

They were checking _just_ openzeppelin library code (AccessControl.sol). IMO it does not make sense to test libraries we are adding to our implementations, much less if they come from OZ. I kept tests where actual parts of our code are checked (ie. PAUSER_ROLE, if it was correctly applied where it should).

packages/solidity-contracts/protocol/utils.ts
DefiCake: To my knowledge they can be merged, but ESLint will complain about it. It was a rule introduced by @luizstacio if I recall correctly, it might have something to do with how the monorepo packages the typings etc.

packages/solidity-contracts/protocol/utils.ts
DefiCake: Done, resolving

Braqzen (approved):

PR #82, opened 2023-10-02
Bridge handshake protocol

DefiCake:
This PR adds a handshake protocol that ensures no funds are lost when depositing in the canonical bridge.


packages/solidity-contracts/contracts/messaging/gateway/v2/FuelBridgeBaseV2.sol
```diff
@@ -0,0 +1,26 @@
+// SPDX-License-Identifier: MIT
```
K1-R1, line 1:
```suggestion
// SPDX-License-Identifier: Apache-2.0
```

K1-R1 (changes requested):
Looking good, only requested changes to the .sol licences

DefiCake:
Rebased with main, re-requesting reviews

K1-R1 (approved):

PR #87, opened 2023-10-09
Use own fixture for ERC721Gatway tests

DefiCake:
This is part of efforts to implement the handshake protocol.

The change decouples the unit testing from the integration testing, which are sharing `harness.ts`, and influences in upgrading the contracts with new features.

Additionally removes some lingering erc20 checks, which do not make sense to keep since we are talking about the erc721 context


K1-R1 (approved):
lgtm

PR #88, opened 2023-10-12
Add SRC-7

### LINKED ISSUE #61: Make the bridge compliant with SRC-7
https://github.com/FuelLabs/sway-standards/issues/21

DefiCake:
Closes #61


K1-R1 (approved):
- Might be time to change `FRC20` to `SRC20` (can be in a separate issue) now that the latter has been chosen for the standard
 -  Was there a reason you decided not to import the standard from the standards repo? If not it's probably the better approach. This approach can be taken when implementing SRC-8 unless you think it would be easier on this PR.

DefiCake:
@K1-R1 

> * Might be time to change `FRC20` to `SRC20` (can be in a separate issue) now that the latter has been chosen for the standard
> * Was there a reason you decided not to import the standard from the standards repo? If not it's probably the better approach. This approach can be taken when implementing SRC-8 unless you think it would be easier on this PR.

I 'll open another PR next to this one to change it to SRC20, maybe we can apply the same as you suggest in your second point.

I just didn't think about using the standards repo, but anyway I have tried and I could not get it to compile, I 'll discuss it with the sway team, it seems like there is something wrong in the standards repo

PR #97, opened 2023-11-02
Add ETH deposit cap

DefiCake:


packages/solidity-contracts/contracts/fuelchain/v2/FuelMessagePortalV2.sol
```diff
@@ -0,0 +1,145 @@
...
+    /// @param recipient The message receiver address or predicate root
+    /// @param data The message data to be sent to the receiver
+    function _sendOutgoingMessage(bytes32 recipient, bytes memory data) internal virtual override {
+        bytes32 sender = bytes32(uint256(uint160(msg.sender)));
+        unchecked {
+            //make sure data size is not too large
+            if (data.length >= MAX_MESSAGE_DATA_SIZE) revert MessageDataTooLarge();
+
+            // v2: increase deposited amount for sender
+            // Do not think this needs an overflow check, you cannot get that much ether
```
Braqzen, line 52: From what I recall some version of solidity 0.8.0+ has built in overflow / underflow checks or something like that

packages/solidity-contracts/contracts/fuelchain/v2/FuelMessagePortalV2.sol
```diff
@@ -0,0 +1,145 @@
...
+                if (amount > type(uint64).max) revert AmountTooBig();
+            }
+
+
+            //emit message for Fuel clients to pickup (messageID calculated offchain)
+            uint nonce = _outgoingMessageNonce;
+            emit MessageSent(sender, recipient, nonce, uint64(amount), data);
+
+            // increment nonce for next message
+            _outgoingMessageNonce = nonce + 1;
```
Braqzen, line 79: would `_outgoingMessageNonce += 1;` cause a read and write or just a write? if it doesn't read then we could remove the local `nonce` above to reduce on costs a little

packages/solidity-contracts/contracts/fuelchain/v2/FuelMessagePortalV2.sol
```diff
@@ -0,0 +1,145 @@
...
+            }
+            
+            //make sure amount fits into the Fuel base asset decimal level
+            uint256 amount = msg.value / PRECISION;
+            if (msg.value > 0) {
+                if (amount * PRECISION != msg.value) revert AmountPrecisionIncompatibility();
+                if (amount > type(uint64).max) revert AmountTooBig();
+            }
+
+
```
Braqzen, line 73:
```suggestion
            // v2: increase global deposited ether
            uint256 globalDepositedAmount = totalDeposited += msg.value;
            if(globalDepositedAmount > depositLimitGlobal) {
                revert GlobalDepositLimit();
            }
            
            //make sure amount fits into the Fuel base asset decimal level
            uint256 amount = msg.value / PRECISION;
            if (msg.value > 0) {
                if (amount * PRECISION != msg.value) revert AmountPrecisionIncompatibility();
                if (amount > type(uint64).max) revert AmountTooBig();
            }

            depositedAmounts[msg.sender] = userDepositedAmount;

```

Save the user some gas in case subsequent checks revert

packages/solidity-contracts/contracts/fuelchain/v2/FuelMessagePortalV2.sol
```diff
@@ -0,0 +1,145 @@
...
+    function _executeMessage(bytes32 messageId, Message calldata message) internal virtual override nonReentrant {
+        if (_incomingMessageSuccessful[messageId]) revert AlreadyRelayed();
+
+        //set message sender for receiving contract to reference
+        _incomingMessageSender = message.sender;
+
+        // v2: update accounting if the message carries an amount 
+        bool success;
+        bytes memory result;
+        if(message.amount > 0) {
```
Braqzen, line 95:
```suggestion
        if (message.amount > 0) {
```

adding a solidity linter would be nice

packages/solidity-contracts/contracts/fuelchain/v2/FuelMessagePortalV2.sol
```diff
@@ -0,0 +1,145 @@
...
+            uint256 withdrawnAmount = message.amount * PRECISION;
+            address recipient = address(uint160(uint256(message.recipient)));
+            uint256 depositedAmount = depositedAmounts[recipient];
+
+            if(depositedAmount < withdrawnAmount) {
+                depositedAmounts[recipient] = 0;
```
Braqzen, line 101: Not sure I understand, are they trying to withdraw more than they have deposited? if that it true then this should error instead of letting them withdraw more

Braqzen (commented):
Haven't really looked through the tests. Wondering if we could abstract away common test utils from the set up

packages/solidity-contracts/contracts/fuelchain/v2/FuelMessagePortalV2.sol
DefiCake: It does an `sload`, that's why it is using a local variable

packages/solidity-contracts/contracts/fuelchain/v2/FuelMessagePortalV2.sol
DefiCake: You could be withdrawing to a different address, not the one you used originally to deposit.

I thought to myself that it might be worth it to completely remove this and not do accounting on withdrawal (save the global counter), just because of this issue, so that we save some gas. But I decided to do the naive implementation first

packages/solidity-contracts/contracts/fuelchain/v2/FuelMessagePortalV2.sol
```diff
@@ -0,0 +1,145 @@
...
+    error GlobalDepositLimit();
+
+    uint public depositLimitGlobal;
+    uint public depositLimitPerAccount;
+
+    uint public totalDeposited;
+    mapping(address => uint) public depositedAmounts;
+
+
+    function initializeV2(
```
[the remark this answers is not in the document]
DefiCake: > more trust because we cannot change things later for whatever reason

Contract being upgradable defeats that. As for the rest of the points, I think it might have some issues with our current CI / CD process. Tagging @luizstacio : you were having problems with immutables / constants, right?

packages/solidity-contracts/contracts/fuelchain/v2/FuelMessagePortalV2.sol
```diff
@@ -0,0 +1,145 @@
...
+        depositLimitGlobal = limit;
+    }
+
+    function setPerAccountDepositLimit(uint limit) external virtual payable onlyRole(DEFAULT_ADMIN_ROLE) {
+        depositLimitPerAccount = limit;
+    }
+
+    function rescueETH(uint amount) external virtual payable onlyRole(DEFAULT_ADMIN_ROLE) {
+        payable(msg.sender).transfer(amount);
+    }
```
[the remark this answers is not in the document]
DefiCake: @Voxelot do we need a rescue function as part of the training wheels requirement?

packages/solidity-contracts/contracts/fuelchain/v2/FuelMessagePortalV2.sol
DefiCake: Indeed, this comment is there because it exists inside an `unchecked` block (look L47), which disables these checks. Will resolve this

K1-R1 (dismissed):
LGTM in it's current state. Will drop an approve here as open discussions only regard removal.

packages/solidity-contracts/contracts/fuelchain/v2/FuelMessagePortalV2.sol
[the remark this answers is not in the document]
DefiCake: Solved by making limits immutables

packages/solidity-contracts/contracts/fuelchain/v2/FuelMessagePortalV2.sol
DefiCake: Resolving

packages/solidity-contracts/contracts/fuelchain/v2/FuelMessagePortalV2.sol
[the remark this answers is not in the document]
DefiCake: Tentatively removed as this functionality can be easily readded and tested

Braqzen (approved):

packages/solidity-contracts/contracts/fuelchain/v2/FuelMessagePortalV2.sol
Voxelot: hmm good point, maybe instead of a per account limit we should just cap the amount for each deposit to rate limit, and just stick to the global limit 🤔

K1-R1 (approved):

packages/solidity-contracts/contracts/fuelchain/v2/FuelMessagePortalV2.sol
DefiCake: I am merging this one and opening another PR to showcase how the alternatives would look, accompanied with docs for the possible scenarios a user might face

PR #100, opened 2023-11-06
Add ERC20 deposit caps

### LINKED ISSUE #58: Training Wheels
Goal: We'd like to be able to deploy future testnet releases to mainnet, while doing everything we can to minimize any potential loss of funds for users due to upgrades, bugs, or other potential issues. This effort would be considered as adding "training wheels" to the bridge that could be removed once the network is stabilized and sufficiently tested.

High level solution:
Value cap the bridge such that users can only bring de minimis amounts of mainnet assets in the early stages. This will make it clear to users that the network is not stabilized yet, and in the case where funds cannot be easily recovered there would be no significant impacts. This could entail whitelisting which tokens are bridgeable, and placing a limit on each kind of asset based on an estimated fair market value. In addition to a global value restriction, there could also be account level restrictions to ensure each user is protected.

DefiCake:
Closes #58


packages/solidity-contracts/test/behaviors/index.ts
Braqzen, line 1: Why do we have multiple versions?

Braqzen (approved):

K1-R1 (approved):

PR #101, opened 2023-11-15
Remove per-account deposit limits

DefiCake:


DefiCake:
@Voxelot here is the alternative with rate limit (per tx) instead of account limit

Braqzen (approved):
Do you think we could update the state prior to the calls being made?

DefiCake:
@Braqzen  

> Do you think we could update the state prior to the calls being made?

Yes and no

- Not with the current approach: the system is designed to allow a lot of freedom wrt how to send messages to other contracts (see next point). It is not really an issue because the _executeMessage has the nonReentrant modifier. Aside of that the current approach is also quite gas intensive.
- Yes, if we implemented an alternative that is more restricted, based on fixed callbacks. Atm the fuel layer can send any kind of message and invoke virtually any method in any contract on the L1. When it comes to message passing what I have usually seen is a generic callback method that the recipient of the message has to implement. (think instead of having a `finalizeWithdrawal method on the erc gateways, we could have a `receiveFuelPortalMessage`). 

It is a trade off. I am personally more comfortable with the second approach, but I see the advantages of the current one.

PR #102, opened 2023-11-15
Replace FRC20 with SRC20

DefiCake:


packages/fungible-token/bridge-fungible-token/Forc.toml
```diff
@@ -7,3 +7,4 @@ name = "bridge_fungible_token"
 [dependencies]
 contract_message_receiver = { path = "../../message-predicates/contract-message-receiver" }
 reentrancy = { git = "https://github.com/FuelLabs/sway-libs", tag = "v0.12.0" }
+src_20 = { git = "https://github.com/FuelLabs/sway-standards", tag = "v0.1.2" }
```
Braqzen, line 10:
```suggestion
src_20 = { git = "https://github.com/FuelLabs/sway-standards", tag = "v0.1.2" }

```

packages/fungible-token/bridge-fungible-token/src/interface/bridge.sw
```diff
@@ -1,5 +1,7 @@
 library;
 
+use std::string::String;
```
K1-R1, line 3: I dont think this is used in the file
```suggestion
```

packages/fungible-token/bridge-fungible-token/tests/functions/src20/mod.rs
```diff
@@ -0,0 +1,98 @@
...
+        interface::src20::{decimals, name, symbol, total_supply},
+        setup::{
+            create_msg_data, create_wallet, get_asset_id, relay_message_to_contract,
+            setup_environment, BridgeFungibleTokenContract, BridgingConfig,
+        },
+    };
+    use fuels::accounts::wallet::WalletUnlocked;
+
+    /// This setup mints tokens so that they are registered as minted assets in the bridge
+    async fn setup_test() -> BridgeFungibleTokenContract<WalletUnlocked> {
```
K1-R1, line 15: Can you move this to utils?

packages/fungible-token/bridge-fungible-token/tests/functions/src20/mod.rs
```diff
@@ -0,0 +1,98 @@
...
+            utxo_inputs.message[0].clone(),
+            utxo_inputs.contract,
+            &utxo_inputs.coin[..],
+        )
+        .await;
+
+        contract
+    }
+
+    #[ignore]
```
K1-R1, line 53: This test is ignored? 
It looks like the comment about the SDK is outdated and so this test can be un-ignored and the assertion added back in?

packages/fungible-token/bridge-fungible-token/src/interface/bridge.sw
```diff
@@ -1,5 +1,7 @@
...
+
```
Braqzen, line 4:
```suggestion
```

Why was this added?

packages/fungible-token/bridge-fungible-token/src/main.sw
```diff
@@ -197,27 +198,32 @@ impl Bridge for Contract {
     }
 }
 
-impl FRC20 for Contract {
+impl SRC20 for Contract {
     #[storage(read)]
-    fn total_supply() -> U256 {
-        let minted: u64 = storage.tokens_minted.try_read().unwrap_or(0);
+    fn total_assets() -> u64 {
```
Braqzen, line 203: Why are the assets and total supply 0 / None?

packages/fungible-token/bridge-fungible-token/src/main.sw
```diff
@@ -257,3 +263,8 @@ fn _asset_to_sub_id(asset_id: AssetId) -> b256 {
     require(sub_id.is_some(), BridgeFungibleTokenError::AssetNotFound);
     sub_id.unwrap()
 }
+
+#[storage(read)]
+fn _check_asset_is_minted_by_contract(asset_id: AssetId) {
+    require(storage.asset_to_sub_id.get(asset_id).try_read().is_some(), BridgeFungibleTokenError::AssetNotFound);
+}
```
Braqzen, line 270: I would remove this because it's only used twice and it has a single require. If it did a bit more then I would potentially keep it

packages/fungible-token/bridge-fungible-token/tests/functions/src20/mod.rs
```diff
@@ -0,0 +1,98 @@
...
+    #[tokio::test]
+    async fn check_total_supply() {
+        // Lacking SDK support on version 0.43
```
Braqzen, line 56: SDK is way ahead of this version

packages/fungible-token/bridge-fungible-token/tests/functions/src20/mod.rs
```diff
@@ -0,0 +1,98 @@
...
+            &utxo_inputs.coin[..],
+        )
+        .await;
+
+        contract
+    }
+
+    #[ignore]
+    #[tokio::test]
+    async fn check_total_supply() {
```
Braqzen, line 55: total_assets test?

packages/fungible-token/bridge-fungible-token/tests/functions/src20/mod.rs
DefiCake: Done

packages/fungible-token/bridge-fungible-token/tests/functions/src20/mod.rs
DefiCake: Fixed

packages/fungible-token/bridge-fungible-token/src/main.sw
DefiCake: FIxed

packages/fungible-token/bridge-fungible-token/src/main.sw
DefiCake: Removed

packages/fungible-token/bridge-fungible-token/tests/functions/src20/mod.rs
DefiCake: Outdated, resolved

packages/fungible-token/bridge-fungible-token/tests/functions/src20/mod.rs
DefiCake: Done

packages/fungible-token/bridge-fungible-token/src/main.sw
```diff
@@ -201,28 +203,45 @@ impl Bridge for Contract {
...
-        None
+    fn total_supply(asset: AssetId) -> Option<u64> {
+        if storage.asset_to_sub_id.get(asset).try_read().is_none() {
+            return None;
+        }
+
+        // Reasoning: an extreme supply token such as SHIBA could overflow the
+        // total supply, which is defined by SRC20 as u64.
+        // .as_u64() will revert on overflow, in that case, it will return u64::max() 
+        Some(storage.tokens_minted.get(asset).read().as_u64().unwrap_or(u64::max()))
```
Braqzen, line 218: So this will technically be an incorrect value under an overflow. Unsure about this

packages/fungible-token/bridge-fungible-token/tests/functions/bridge/mod.rs
```diff
@@ -16,6 +16,7 @@ mod success {
 
     use super::*;
 
+    use crate::utils::interface::src20::total_supply;
```
Braqzen, line 19: I would always merge imports.

What happens if you run `cargo clippy --all-features --all-targets -- -D warnings`?

Braqzen (commented):
@bitzoic what do you think about the SRC20 here? It's a little different

packages/fungible-token/bridge-fungible-token/tests/functions/bridge/mod.rs
DefiCake: Complains about other things, not this one specifically

```
   Compiling fuel-contract-message-predicate v0.3.0 (/Users/mad/fuel/bridge/packages/message-predicates/contract-message-predicate)
    Checking bridge-fungible-token-contract v0.3.0 (/Users/mad/fuel/bridge/packages/fungible-token/bridge-fungible-token)
error: `assert!(false, ..)` should probably be replaced
   --> packages/fungible-token/bridge-fungible-token/tests/functions/bridge/mod.rs:347:17
    |
347 |                 assert!(false, "Transaction did not succeed")
    |                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = help: use `panic!(..)` or `unreachable!(..)`
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#assertions_on_constants
    = note: `-D clippy::assertions-on-constants` implied by `-D warnings`

error: `assert!(false, ..)` should probably be replaced
    --> packages/fungible-token/bridge-fungible-token/tests/functions/message_receiver/mod.rs:1185:17
     |
1185 |                 assert!(false, "Transaction did not revert");
     |                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
     |
     = help: use `panic!(..)` or `unreachable!(..)`
     = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#assertions_on_constants
```

My go to command with clippy is `cargo clippy --all-features --all-targets --fix`, I 'll make sure to run this one as well.

I 'll push a commit to merge this one, and fix the others that -D warnings gives too

DefiCake: Fixed

packages/fungible-token/bridge-fungible-token/src/main.sw
DefiCake: Fixed: it now registers a refund when the total_supply cap is reached

packages/fungible-token/bridge-fungible-token/src/main.sw
```diff
@@ -52,7 +57,8 @@ configurable {
 storage {
     asset_to_sub_id: StorageMap<AssetId, b256> = StorageMap {},
```
bitzoic, line 58: I see this is only used in the withdrawal function of the bridge, wouldn't it make more sense to pass it as an argument to the `withdraw(to: b256, sub_id: SubId)` function and assert that the `msg_asset_id() == AssetId::new(contract_id(), sub_id)`?

packages/fungible-token/bridge-fungible-token/src/main.sw
```diff
@@ -197,27 +214,46 @@ impl Bridge for Contract {
...
     #[storage(read)]
-    fn name() -> str[64] {
-        NAME
+    fn name(asset: AssetId) -> Option<String> {
+        if storage.asset_to_sub_id.get(asset).try_read().is_none() {
+            return None;
+        }
+
+        Some(String::from_ascii_str(from_str_array(NAME)))
     }
```
bitzoic, line 239:
```suggestion
    fn name(asset: AssetId) -> Option<String> {
        match storage.total_supply.get(asset).try_read() {
            Some(v) => Some(String::from_ascii_str(from_str_array(NAME))),
            None => None,
        }        
    }
```

bitzoic (commented):
With the SRC-20 standard you no longer need the empty spaces at the end of your configurable for `NAME` and `SYMBOL`. If every bridged token is using the same name then just "MyToken" or "BridgedAsset" will suffice. I'm also unsure if every asset bridged uses the same decimals i.e. NFTs and high supply assets but I don't know the specific implementation that is happening here.

packages/fungible-token/bridge-fungible-token/src/main.sw
```diff
@@ -88,6 +92,20 @@ impl MessageReceiver for Contract {
...
+
+                let current_total_supply = storage.tokens_minted.get(asset_id).try_read().unwrap_or(0);
+                let new_total_supply = current_total_supply + amount;
+
+                if new_total_supply < current_total_supply {
+                    register_refund(message_data.from, message_data.token_address, message_data.token_id, message_data.amount);
+                    return;
+                }
+
+                enable_panic_on_overflow();
```
Braqzen, line 105: Wondering what happens if we return above but do not reach this line. Do we continue to have disabled panics for overflow or are they re-enabled automatically in the subsequent calls?

packages/fungible-token/bridge-fungible-token/src/main.sw
DefiCake: This would imply that callers have to be aware of the `sub_id` of the asset they are withdrawing. There has been a lot of discussion around this and I think we have not yet reached an agreement, as you said, it might be that the `sub_id` might be available as part of the message in the future. For the time being, and given that block space remains cheap, the consensus is that we keep the mapping of asset_id <-> sub_id in storage so that callers do not need to be aware of the sub_id.

I do not have a strong opinion on this though.

packages/fungible-token/bridge-fungible-token/src/main.sw
DefiCake: To the best of my knowledge, this function cannot be called from other contexts inside the contract (it acts as an external function in solidity). 

OTOH, I 'd assume that contract-to-contract calls cannot "poison" these kind of flag configurations.

packages/fungible-token/bridge-fungible-token/src/main.sw
DefiCake: I 've also refactored the use of `b256` types and replaced it with SubId and AssetId where needed

packages/fungible-token/bridge-fungible-token/tests/functions/message_receiver/mod.rs
```diff
@@ -120,19 +121,244 @@ mod success {
...
+            amount,
+            token_address,
+            from,
+            token_id: _,
+        } = refund_registered_events[0];
+
+        // Check logs
+        assert_eq!(amount, Bits256(encode_hex(max_deposit_amount)));
+        assert_eq!(token_address, Bits256::from_hex_str(BRIDGED_TOKEN).unwrap());
+        assert_eq!(from, Bits256::from_hex_str(FROM).unwrap());
```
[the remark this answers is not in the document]
DefiCake: Does not hurt to do so 👍🏻  I just maintained consistency across tests

packages/fungible-token/bridge-fungible-token/src/main.sw
```diff
@@ -192,32 +209,46 @@ impl Bridge for Contract {
...
-        U256::from((0, 0, 0, minted))
+    #[storage(read)]
+    fn total_supply(asset: AssetId) -> Option<u64> {
+        storage.tokens_minted.get(asset).try_read()
     }
 
     #[storage(read)]
-    fn name() -> str[64] {
-        NAME
+    fn name(asset: AssetId) -> Option<String> {
```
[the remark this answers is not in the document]
DefiCake: For context, let's remember that an instance of this sway contract can only hold one EVM token, 
- with ERC20 tokens using a token_id = sub_id = 0, which translates as having an unique AssetId that can have any amount from 1 to U64.max
- with ERC721 using token_id for each collectible, which translates as having multiple AssetIds that can have only 1 unit each, but still all belong to the same "collection" that is ERC721, and all of these have the same symbol

Illustration:
- ERC20: you are bridging 10 USDT and that translates to a single AssetId say 0x0..abcd, with amount 10. Then, if you query the name for AssetId 0x0..abcd, it will return "Fuel USDT"
- ERC721: you are bridging Ape 1, Ape 2 and Ape 3 with token_id's 1, 2, 3. That translates to AssetId 0x0..1, 0x0..2, 0x0..3, with amount 1 for each one of them. Then, if you query any of these assets for its name, it should return the same name: "Fuel Ape", which is the original ERC721 collection.

Braqzen (approved):

PR #110, opened 2023-12-11
Add beta 5 deployment artifacts

DefiCake:


Braqzen (approved):
utACK. I have not double checked any of the addresses or integers

PR #112, opened 2023-12-14
ci: add pnpm audit and update vulnerabilities

luizstacio:


packages/solidity-contracts/package.json
```diff
@@ -42,11 +42,11 @@
     "@ethersproject/bytes": "^5.7.0",
     "@ethersproject/providers": "^5.7.0",
     "@fuel-contracts/merkle-sol": "^0.1.4",
-    "@nomicfoundation/hardhat-chai-matchers": "^1.0.6",
+    "@nomicfoundation/hardhat-chai-matchers": "^2.0.2",
```
DefiCake, line 45: chai matchers v2 only work with ethers v6 - if you need to carry on with this, let me know and I 'll take care of the changes, as they expand to a lot of avenues (one of them is the typechain artifacts, although I think you are using viem in the frontend so I guess that would not be an issue)

DefiCake:
Replaced by #126

PR #119, opened 2024-01-22
Bump forc

DefiCake:
This PR bumps forc to 0.49.1.

As part of the process the following was also needed:
- Revert back the github actions to use the orignal `forc` binary instead of the one loaded with the `ts-sdk`. This is because otherwise, we are stuck with the forc binary that comes with the sdk, whereas we need to remain mobile
- Apply formatting to appease the new forc rules

### DIFF
--- a/.changeset/chilled-candles-explain.md
+++ b/.changeset/chilled-candles-explain.md
@@ -1,0 +1,5 @@
+---
+'@fuel-bridge/test-utils': patch
+---
+
+feat: update fuels-ts to 0.71.1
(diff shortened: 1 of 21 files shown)


K1-R1 (approved):
I'm not that familiar with the `.changeset` files; should there just be a single one, or is it fine for there to be multiple?

DefiCake:
> I'm not that familiar with the `.changeset` files; should there just be a single one, or is it fine for there to be multiple?

Not sure if there 'd be a situation where it is fine for there to be multiple, but not this one. I removed the spurious one

K1-R1 (approved):

PR #124, opened 2024-02-08
Bump forc to 0.50.0

DefiCake:


packages/fungible-token/bridge-fungible-token/src/cast.sw
```diff
@@ -6,9 +6,11 @@ impl From<b256> for u256 {
     fn from(value: b256) -> Self {
         value.as_u256()
     }
+}
 
-    fn into(self) -> b256 {
-        let result: b256 = asm(r1: self) {
+impl From<u256> for b256 {
```
K1-R1, line 11: Nothing wrong with this; but sway-lib-core provides `.as_b256()` which may be a better choice for maintainability: https://github.com/FuelLabs/sway/blob/304317cc4ad4bc9e5600c9c7b001f918abe2fcee/sway-lib-core/src/primitive_conversions.sw#L273

K1-R1 (approved):
LGTM. I'll leave it to author's choice as to implement the comment below

bitzoic (approved):

packages/fungible-token/bridge-fungible-token/src/cast.sw
DefiCake: Noted it as an issue 👍🏻

PR #126, opened 2024-02-15
Add pnpm audit to CI

DefiCake:


packages/test-utils/src/utils/ethers/getOrDeployERC721Contract.ts
```diff
@@ -6,23 +6,24 @@ import type { TestEnvironment } from '../setup';
...
     const eth_tokenFactory = new NFT__factory(ethDeployer);
-    ethTestNft = await eth_tokenFactory.deploy();
-    await ethTestNft.deployed();
+    ethTestNft = await eth_tokenFactory
+      .deploy()
+      .then((tx) => tx.waitForDeployment());
+
     debug(
-      `Ethereum ERC-721 token contract created at address ${ethTestNft.address}.`
+      `Ethereum ERC-721 token contract created at address ${await ethTestNft.getAddress()}.`
```
luizstacio, line 21: Wouldn't make sense to add the value a const to avoid two async calls? I saw other parts with a similar pattern that could also be fixed.

luizstacio (changes requested):

packages/test-utils/src/utils/ethers/getOrDeployERC721Contract.ts
DefiCake: I have reviewed occurrences of `getAddress()` and cached the value where possible. it would be possible to eliminate more of these, but a deeper refactor would be required, and the gains are not very big. Most of this `getAddress()` calls are async in case the value is not found locally, but in reality most of these are already locally stored and thus even if the call is async, it does not do any network request.

PR #143, opened 2024-03-06
Remove b256-u256 conversions

### LINKED ISSUE #125: Use as_b256() for u256 -> b256 conversions

DefiCake:
Closes #125 

Took the opportunity and removed all unneeded conversions now that we can use `u256` at storage.


bitzoic (approved):
LGTM

ZERO_U256 would be a nice addition to the std-lib

PR #178, opened 2024-05-02
Fix Hexens FUEL1-10: use custom error for unknown relay reverts

DefiCake:
This PR changes FuelMessagePortal 's revert on relay (`revert("Message relay failed")`) to a custom error to achieve some gas savings

### DIFF
--- a/packages/solidity-contracts/contracts/fuelchain/FuelMessagePortal/v3/FuelMessagePortalV3.sol
+++ b/packages/solidity-contracts/contracts/fuelchain/FuelMessagePortal/v3/FuelMessagePortalV3.sol
@@ -10,6 +10,7 @@ contract FuelMessagePortalV3 is FuelMessagePortalV2 {
 
     error MessageBlacklisted();
     error WithdrawalsPaused();
+    error MessageRelayFailed();
 
     bool public withdrawalsPaused;
     mapping(bytes32 => bool) public messageIsBlacklisted;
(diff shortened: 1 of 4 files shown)


packages/solidity-contracts/contracts/fuelchain/FuelMessagePortal/v3/FuelMessagePortalV3.sol
```diff
@@ -93,6 +94,52 @@ contract FuelMessagePortalV3 is FuelMessagePortalV2 {
...
+                assembly {
+                    let returndata_size := mload(result)
+                    revert(add(32, result), returndata_size)
+                }
+            }
+            revert MessageRelayFailed();
+        }
+
+        //unset message sender reference
+        _incomingMessageSender = NULL_MESSAGE_SENDER;
```
K1-R1, line 134: What's the benefit of resetting here?

K1-R1 (approved):

packages/solidity-contracts/contracts/fuelchain/FuelMessagePortal/v3/FuelMessagePortalV3.sol
DefiCake: When a messag from L2 to L1 is relayed, this parameter is set to the sender temporarily.

Then a possible callback is made to a L1 contract (e.g., call the L1 bridge to release tokens)

The L1 called contract has the option to query the portal for who is the sender of the message (e.g., the L1 bridge queries the message portal to ensure that the incoming message sender is the L2 counterpart)

After L1 target contract has been called, we reset the variable. For the flow of the bridge itself this might not be strictly needed as we only accept one message at a time via callback, but there might other flows and implementations that do need to validate the correctness of this variable at a time other than the callback.

PR #204, opened 2024-06-05
feat: make FuelChainState params configurable on deployment

### LINKED ISSUE #203: Make FuelChainState params configurable on deployment

DefiCake:
Closes #203


K1-R1 (approved):

packages/solidity-contracts/contracts/fuelchain/FuelChainState.sol
```diff
@@ -66,7 +69,11 @@ contract FuelChainState is Initializable, PausableUpgradeable, AccessControlUpgr
 
     /// @notice Constructor disables initialization for the implementation contract
     /// @custom:oz-upgrades-unsafe-allow constructor
-    constructor() {
+    constructor(uint256 timeToFinalize, uint256 blocksPerCommitInterval) {
+        TIME_TO_FINALIZE = timeToFinalize;
+        COMMIT_COOLDOWN = uint32(timeToFinalize) * 8;
```
BarneyChambers, line 74: what's the reason for the COMMIT_COOLDOWN to be 8 times the `timeToFinalize`?

DefiCake: It is arbitrary.

As it was constant before, it was meant to be 24 hours (3 hours time to finalize * 8 = 24h).

On a second thought it should be configurable too, as depending on the config you could be setting the cooldown beyond the number of commit slots. That could be an issue.

packages/solidity-contracts/deploy/hardhat/001.chain_state.ts
```diff
@@ -3,6 +3,9 @@ import type { DeployFunction } from 'hardhat-deploy/dist/types';
 
 import { FuelChainState__factory as FuelChainState } from '../../typechain';
 
+const BLOCKS_PER_COMMIT_INTERVAL = 30;
+const TIME_TO_FINALIZE = 5;
```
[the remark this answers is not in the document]
DefiCake: Not necessarily. The commit cooldown refers to a slot in `Commit[NUM_COMMIT_SLOTS] private _commitSlots;`, meaning that you cannot write to _commitSlots[0] more than once between cooldowns, but you can definitely push to index 1, 2, 3, 4... as much as you want

packages/solidity-contracts/contracts/fuelchain/FuelChainState.sol
BarneyChambers: Ok cool, that makes sense! I agree that it should also be configurable

K1-R1 (approved):

PR #229, opened 2024-08-09
chore: add full e2e dockerfile and upgrade node version

DefiCake:


K1-R1 (approved):

viraj124 (approved):

packages/integration-tests/tests/bridge_erc20.ts
```diff
@@ -84,11 +84,17 @@ describe('Bridging ERC20 tokens', async function () {
 
     // mint tokens as starting balances
 
-    await eth_testToken.mint(await env.eth.deployer.getAddress(), 10_000);
+    await eth_testToken
+      .mint(await env.eth.deployer.getAddress(), 10_000)
```
viraj124, line 88: nit we can add the num in a var

viraj124 (approved):

packages/integration-tests/tests/bridge_erc20.ts
DefiCake: I think that the future token rate limit PR will be a better place to do this, will keep an eye on it

Issue #210, opened 2024-06-10
(Latacora) LT-APP-1: Implement on-chain rate limiting for withdrawals


DefiCake:

DefiCake:
Reopening, erc20 rate limiter not implemented yet

DefiCake:
No, we keep this one so that Latacora can have a single point where they can see the relevant commits

Issue #173, opened 2024-04-24
Tight pack deposit info


DefiCake:
At FuelERC20GatewayV4, it is possible to pack this information:

```sol
mapping(address => uint256) internal _deposits;
mapping(address => uint256) internal _depositLimits;
mapping(address => uint256) internal _decimalsCache;
```

Under the a structure to leverage tight packing

PR #239, opened 2024-08-20
ERC20 rate limiter

viraj124:
- closes https://github.com/FuelLabs/fuel-bridge/issues/210
- adds rate limit feature that can be set individually for each erc20 token that the bridge supports

### DIFF
--- a/packages/solidity-contracts/contracts/messaging/gateway/FuelERC20Gateway/FuelERC20GatewayV4.sol
+++ b/packages/solidity-contracts/contracts/messaging/gateway/FuelERC20Gateway/FuelERC20GatewayV4.sol
@@ -34,6 +34,7 @@ contract FuelERC20GatewayV4 is
     error GlobalDepositLimit();
     error CannotDepositZero();
     error CannotWithdrawZero();
+    error InvalidAssetIssuerID();
     error InvalidSender();
     error InvalidAmount();
     error RateLimitExceeded();
@@ -206,6 +207,8 @@ contract FuelERC20GatewayV4 is
     /// @param amount Amount of tokens to deposit
     /// @dev Made payable to reduce gas costs
     function deposit(bytes32 to, address tokenAddress, uint256 amount) external payable virtual whenNotPaused {
+        if (assetIssuerId == bytes32(0)) revert InvalidAssetIssuerID();
+
         uint8 decimals = _getTokenDecimals(tokenAddress);
         uint256 l2MintedAmount = _adjustDepositDecimals(decimals, amount);
 
@@ -234,6 +237,8 @@ contract FuelERC20GatewayV4 is
         uint256 amount,
         bytes calldata data
     ) external payable virtual whenNotPaused {
+        if (assetIssuerId == bytes32(0)) revert InvalidAssetIssuerID();
+
         uint8 decimals = _getTokenDecimals(tokenAddress);
         uint256 l2MintedAmount = _adjustDepositDecimals(decimals, amount);
 
@@ -252,6 +257,8 @@ contract FuelERC20GatewayV4 is
     }
 
     function sendMetadata(address tokenAddress) external payable virtual whenNotPaused {
+        if (assetIssuerId == bytes32(0)) revert InvalidAssetIssuerID();
+
         bytes memory metadataMessage = abi.encodePacked(
             assetIssuerId,
             uint256(MessageType.METADATA),
--- a/.changeset/witty-eyes-march.md
+++ b/.changeset/witty-eyes-march.md
@@ -1,0 +1,5 @@
+---
+'@fuel-bridge/solidity-contracts': major
+---
+
+add zero address check for asset issuer id
--- a/packages/solidity-contracts/test/behaviors/erc20GatewayV4.behavior.test.ts
+++ b/packages/solidity-contracts/test/behaviors/erc20GatewayV4.behavior.test.ts
@@ -362,6 +362,28 @@ export function behavesLikeErc20GatewayV4(fixture: () => Promise<Env>) {
             );
           });
 
+          it('reverts when asset issuer id is not set', async () => {
+            const {
+              erc20Gateway,
+              token,
+              signers: [deployer, user],
+            } = env;
+
+            const depositTo = randomBytes32();
+            const depositAmount = MaxUint256;
+
+            await erc20Gateway.connect(deployer).setAssetIssuerId(ZeroHash);
+
+            const tx = erc20Gateway
+              .connect(user)
+              .deposit(depositTo, token, depositAmount);
+
+            await expect(tx).to.be.revertedWithCustomError(
+              erc20Gateway,
+              'InvalidAssetIssuerID'
+            );
+          });
+
           it('calls FuelMessagePortal', async () => {
             const {
               erc20Gateway,


packages/solidity-contracts/contracts/messaging/gateway/FuelERC20Gateway/FuelERC20GatewayV4.sol
```diff
@@ -66,10 +71,22 @@ contract FuelERC20GatewayV4 is
...
-    mapping(address => uint256) internal _depositLimits;
-    mapping(address => uint256) internal _decimalsCache;
-    mapping(bytes32 => bool) internal _isBridge;
+    mapping(address => uint256) _deposits;
+    mapping(address => uint256) _depositLimits;
+    mapping(address => uint256) _decimalsCache;
+    mapping(bytes32 => bool) _isBridge;
+
+    /// @notice Amounts already withdrawn this period for each token.
+    mapping(address => uint256) public rateLimitDuration;
```
DefiCake, line 80: let's the take the chance to implement https://github.com/FuelLabs/fuel-bridge/issues/173

until mainnet launch we won't care if the erc20 gateway breaks backwards compatibility in storage

packages/solidity-contracts/contracts/messaging/gateway/FuelERC20Gateway/FuelERC20GatewayV4.sol
```diff
@@ -121,6 +140,57 @@ contract FuelERC20GatewayV4 is
...
     }
 
+    /**
+     * @notice Intitializing rate limit params for each token. 
+     * @param _token The token address to set rate limit params for.
+     * @param _limitAmount The amount to reset the limit to.
+     * @param _rateLimitDuration rate limit duration.
+     */
+    function initializeRateLimit(address _token, uint256 _limitAmount, uint256 _rateLimitDuration) external onlyRole(SET_RATE_LIMITER_ROLE) {
+        if (currentPeriodEnd[_token] != 0) revert RateLimitAlreadySet();
```
DefiCake, line 150: Why isn't it possible to change the duration after it has been set?

packages/solidity-contracts/contracts/messaging/gateway/FuelERC20Gateway/FuelERC20GatewayV4.sol
```diff
@@ -226,6 +296,9 @@ contract FuelERC20GatewayV4 is
         uint8 decimals = _getTokenDecimals(tokenAddress);
         uint256 amount = _adjustWithdrawalDecimals(decimals, l2BurntAmount);
 
+        // rate limit check
+        _addWithdrawnAmount(tokenAddress, amount);
```
DefiCake, line 300: Wouldn't we achieve significant gas savings if we added a check for the existence of a rate limiter?

packages/solidity-contracts/contracts/messaging/gateway/FuelERC20Gateway/FuelERC20GatewayV4.sol
```diff
@@ -121,6 +140,57 @@ contract FuelERC20GatewayV4 is
...
+        currentPeriodEnd[_token] = block.timestamp + _rateLimitDuration;
+        limitAmount[_token] = _limitAmount;
+    }
+
+    /**
+     * @notice Resets the rate limit amount.
+     * @param _token The token address to set rate limit for.
+     * @param _amount The amount to reset the limit to.
+     */
+    function resetRateLimitAmount(address _token, uint256 _amount) external onlyRole(SET_RATE_LIMITER_ROLE) {
```
DefiCake, line 162: Shouldn't we have events?

packages/solidity-contracts/contracts/messaging/gateway/FuelERC20Gateway/FuelERC20GatewayV4.sol
DefiCake: Ignore the above comment, we are finally going to go with the old storage

packages/solidity-contracts/contracts/messaging/gateway/FuelERC20Gateway/FuelERC20GatewayV4.sol
```diff
@@ -66,10 +74,22 @@ contract FuelERC20GatewayV4 is
...
     bytes32 public assetIssuerId;
 
-    mapping(address => uint256) internal _deposits;
-    mapping(address => uint256) internal _depositLimits;
-    mapping(address => uint256) internal _decimalsCache;
-    mapping(bytes32 => bool) internal _isBridge;
+    mapping(address => uint256) _deposits;
+    mapping(address => uint256) _depositLimits;
+    mapping(address => uint256) _decimalsCache;
+    mapping(bytes32 => bool) _isBridge;
```
PraneshASP, line 80: nit: good to add visibility as `private`

PraneshASP: also we need to decide the naming convention. should we use `_` prefix for private vars or only for params and local vars?

packages/solidity-contracts/contracts/messaging/gateway/FuelERC20Gateway/FuelERC20GatewayV4.sol
PraneshASP: so the workflow would be to call initialzeRateLimit() first then to change the config `resetRateLimit()` must be used?

also good to emit an event here. can be generic as `RateLimitUpdated()` which can be used in both methods.

PraneshASP (commented):
few nits

packages/solidity-contracts/contracts/messaging/gateway/FuelERC20Gateway/FuelERC20GatewayV4.sol
DefiCake: We cannot do that, as we cannot initialize a rate limiter for every token that will ever exist. By default we must use no rate limiter

packages/solidity-contracts/contracts/messaging/gateway/FuelERC20Gateway/FuelERC20GatewayV4.sol
DefiCake: Please make it dynamic. The reasons why it is static in FuelMessagePortal is that we can change it via new deployment (it is a constant / immutable), but here we cannot leverage that advantage, as we need an storage variable for each token. Thus it makes no sense to have that limitation.

packages/solidity-contracts/test/erc721Gateway.test.ts
```diff
@@ -32,7 +32,6 @@ import {
   TIME_TO_FINALIZE,
   COMMIT_COOLDOWN,
 } from './utils';
-
```
DefiCake, line 35: ?

packages/solidity-contracts/contracts/messaging/gateway/FuelERC20Gateway/FuelERC20GatewayV4.sol
```diff
@@ -66,10 +74,22 @@ contract FuelERC20GatewayV4 is
     bool public whitelistRequired;
     bytes32 public assetIssuerId;
 
-    mapping(address => uint256) internal _deposits;
-    mapping(address => uint256) internal _depositLimits;
-    mapping(address => uint256) internal _decimalsCache;
-    mapping(bytes32 => bool) internal _isBridge;
+    mapping(address => uint256) private _deposits;
```
DefiCake, line 77: Why make them private?

DefiCake (approved):
lgtm

PR #249, opened 2024-08-27
Pin npm dependencies

viraj124:
- closes https://github.com/FuelLabs/fuel-bridge/issues/241
- adds `.npmrc` to automate pin future dependencies that are installed


PraneshASP (approved):

DefiCake:
I am concerned that there is a downgrade in the .lock file. Is there a way to easily pin them to the version where they are sitting now?

PR #250, opened 2024-08-27
Add zero address check for asset issuer ID in erc20 gateway v4

viraj124:
- closes https://github.com/FuelLabs/fuel-bridge/issues/231


PraneshASP (dismissed):

DefiCake:
Sorry I might have needed to be more explicit, but we need the check when doing deposits, not when configuring the value.

In fact, a zero value for the setter might be useful to disable deposits altogether

PraneshASP (approved):

PR #267, opened 2024-09-03
use mainnet tokens in e2e erc20 gateway test

viraj124:
- closes https://github.com/FuelLabs/fuel-bridge/issues/227


packages/integration-tests/tests/bridge_erc20.ts
```diff
@@ -141,8 +196,76 @@ describe('Bridging ERC20 tokens', async function () {
...
+    ]);
+    await env.eth.provider.send('hardhat_setCode', [
+      USDC_ADDRESS,
+      runtimeBytecodeUSDCToken,
+    ]);
+    await env.eth.provider.send('hardhat_setCode', [
+      WBTC_ADDRESS,
+      runtimeBytecodeWBTCToken,
+    ]);
+    await env.eth.provider.send('hardhat_setCode', [
```
DefiCake, line 231: Move this to hardhat deploy scripts at `packages/solidity-contracts/deploy/hardhat`, we want this to be available to anyone that runs the envs

packages/integration-tests/tests/bridge_erc20.ts
```diff
@@ -248,112 +412,156 @@ describe('Bridging ERC20 tokens', async function () {
...
-        .then((tx) => tx.wait());
-
-      // parse events from logs
-      const [event, ...restOfEvents] =
-        await env.eth.fuelMessagePortal.queryFilter(
-          env.eth.fuelMessagePortal.filters.MessageSent,
-          receipt.blockNumber,
-          receipt.blockNumber
+      // TODO: ADD WETH
+      for (let i = 0; i < fuelAssetId.length; i++) {
```
DefiCake, line 533: All of these for loops make the testing logic more convoluted than it already is. Ideally avoid for loops inside `it` blocks. 

Try to use a testing table, take example of https://github.com/FuelLabs/fuel-bridge/blob/0b6e3b198707d9fd7000d255b0ad3164d7e64507/packages/solidity-contracts/test/behaviors/erc20GatewayV4.behavior.test.ts#L317

DefiCake (commented):
I support the initiative to expand test coverage, though initially we just had to expose the tokens. Let's take the chance then to improve coverage on these cases

packages/integration-tests/tests/bridge_erc20.ts
DefiCake: if you add it to hardhat deploy scripts, they should be available in e2e tests

DefiCake: We need them in hardhat scripts because other repositories will be using the e2e environment - ie, someone will run our docker images and expect everything to already be deployed and ready.

Why do you need external inputs? afaik, all the information we need to deploy these tokens is static. We know how USDC, USDT and WETH behave.

packages/integration-tests/tests/bridge_mainnet_tokens.ts
```diff
@@ -0,0 +1,768 @@
...
+      fuel_bridgeImpl = implementation;
+
+      fuel_bridgeContractId = fuel_bridge.id.toHexString();
+
+      await env.eth.fuelERC20Gateway.setAssetIssuerId(fuel_bridgeContractId);
+
+      fuelAssetId = getTokenId(fuel_bridge, value);
+
+      // initializing rate limit params for the token
+      if (index == tokenAddresses.length - 1) {
```
DefiCake, line 240: Why not add `18` to the decimals array instead of an if-else block here?

packages/integration-tests/tests/bridge_mainnet_tokens.ts
```diff
@@ -0,0 +1,768 @@
...
+const tokenAddresses: string[] = [
+  USDT_ADDRESS,
+  USDC_ADDRESS,
+  WBTC_ADDRESS,
+  WETH_ADDRESS,
+];
+
+const decimals: bigint[] = [6n, 6n, 8n];
+
+for (const [index, value] of tokenAddresses.entries()) {
```
DefiCake, line 57: I think you can cut the test time if you move this loop lower, so that you don't need to be deploying contracts in each loop, you deploy just once

packages/integration-tests/tests/bridge_mainnet_tokens.ts
```diff
@@ -0,0 +1,768 @@
...
+      // withdraw tokens back to the base chain
+      fuel_bridge.account = fuelTokenSender;
+      const paddedAddress =
+        '0x' + ethereumTokenReceiverAddress.slice(2).padStart(64, '0');
+      const fuelTokenSenderBalance = await fuelTokenSender.getBalance(
+        fuel_AssetId
+      );
+
+      let transactionRequest;
+      if (decimals < 18) {
```
DefiCake, line 97: Withdrawals are really decimals agnostic, can we refactor this?

packages/integration-tests/tests/bridge_mainnet_tokens.ts
```diff
@@ -0,0 +1,608 @@
...
+          fuelTokenReceiverBalance = await fuelTokenReceiver.getBalance(
+            fuelAssetId
+          );
+        });
+
+        it('Bridge ERC20 via FuelERC20Gateway', async () => {
+          // approve FuelERC20Gateway to spend the tokens
+
+          const token =
+            index < tokenAddresses.length - 1 ? customToken : weth_testToken;
```
DefiCake, line 271: Does this `it` block need to segregate between a normal erc20 and a weth?

PraneshASP (approved):

DefiCake (approved):

PR #280, opened 2024-09-05
feat: prepare sway scripts for mainnet

DefiCake:


K1-R1 (approved):

viraj124 (approved):

packages/test-utils/src/scripts/check-balances.ts
```diff
@@ -0,0 +1,30 @@
...
+import { password } from '@inquirer/prompts';
+import { Provider, WalletUnlocked } from 'fuels';
+
+let { L2_ADDRESS, L2_RPC } = process.env;
+
+const main = async () => {
+  const provider = await Provider.create(L2_RPC, { resourceCacheTTL: -1 });
+
+  if (!L2_ADDRESS) {
+    const privKey = await password({ message: 'Enter private key' });
```
PraneshASP, line 14: nit: why get a privatekey instead of address?

packages/test-utils/src/scripts/transfer-bridge-ownership.ts
```diff
@@ -0,0 +1,73 @@
+/**
+ * This is a stand-alone script that upgrades the bridge
```
PraneshASP, line 2: misleading comment.

packages/test-utils/src/scripts/verify-asset-metadata.ts
```diff
@@ -0,0 +1,106 @@
+/**
+ * This is a stand-alone script that
```
PraneshASP, line 2: misleading comment

packages/test-utils/src/scripts/check-balances.ts
DefiCake: So that you can get information about the loaded address from a certain private key

packages/test-utils/src/scripts/verify-asset-metadata.ts
DefiCake: Fixed

PraneshASP (approved):

viraj124 (approved):

PR #318, opened 2024-10-01
fix: reduce FuelERC20GatewayV4 storage gap

DefiCake:


viraj124 (approved):

packages/solidity-contracts/contracts/messaging/gateway/FuelERC20Gateway/FuelERC20GatewayV4.sol
```diff
@@ -449,5 +449,5 @@ contract FuelERC20GatewayV4 is
      * variables without shifting down storage in the inheritance chain.
      * See https://docs.openzeppelin.com/contracts/4.x/upgradeable#storage_gaps
      */
-    uint256[49] private __gap;
+    uint256[48] private __gap;
```
viraj124, line 452: why are we reducing it since previous versions also had `49`

viraj124 (approved):
just one comment besides that lgtm

packages/solidity-contracts/contracts/messaging/gateway/FuelERC20Gateway/FuelERC20GatewayV4.sol
viraj124, line 452: this is a precautionary update since I noticed while upgrading we didn't got errors
also just wanted to check why are we reducing it by 1 as we added 5 mappings from the prev version

PraneshASP (approved):

PR #324, opened 2024-10-03
feat: Use EIP1153 (transient storage) for re-entrancy checks

viraj124:
- closes https://github.com/FuelLabs/fuel-bridge/issues/145
- update solidity compiler version to `0.8.24`


packages/solidity-contracts/contracts/security/ReentrancyGuardTransientUpgradable.sol
```diff
@@ -0,0 +1,81 @@
...
+     * @dev Unauthorized reentrant call.
+     */
+    error ReentrancyGuardReentrantCall();
+
+    uint256 private constant _NOT_ENTERED = 0;
+    uint256 private constant _ENTERED = 1;
+
+    // @notice acts as the storage slot in transient storage
+    // avoiding the use of a namespace storage slot, to avoid storage collisions with upgrades
+    uint256 __transientSlotPlaceholder;
```
DefiCake, line 22:
```suggestion
    uint256 private _status;
```

a varname change will be flagged by the plugin

DefiCake (changes requested):

packages/solidity-contracts/contracts/fuelchain/FuelMessagePortal.sol
```diff
@@ -1,17 +1,17 @@
 // SPDX-License-Identifier: MIT
-pragma solidity 0.8.9;
+pragma solidity 0.8.24;
```
DefiCake, line 2: If I am not mistaken, the reentrancy guard only needs to be applied to the message portal.

Do we need to upgrade the compiler for all the contracts in the code base?

viraj124: we were earlier using the specific version `0.8.9`,  but with `0.8.24` version, we had a choice that, either we upgrade the compiler versions for all contracts to `0.8.24`, or we use the [caret version](https://medium.com/@nareshmmr/why-pragma-solidity-0-x-x-is-important-in-solidity-smart-contracts-9fa669f0110f) which is `^0.8.9` in all contracts so it is compatible with `0.8.24` 

I chose to go with using the specific version

DefiCake: Can you not configure several compilers in the hardhat config? There is a compilers section that accepts an array

viraj124: yeah, I was just trying to avoid using multiple compiler versions, but yeah will do that

DefiCake (approved):
This looks good to me. Please run a manual prepareUpgrade run for these contracts in a tenderly fork before merging

PraneshASP (approved):

PraneshASP (approved):

DefiCake (approved):

PR #336, opened 2024-10-15
feat: add manual ci for querying grant role events

### LINKED ISSUE #337: (feat): add manual ci for querying grant role events

viraj124:
- closes #337

### DIFF
--- a/packages/solidity-contracts/scripts/hardhat/eventFilter.ts
+++ b/packages/solidity-contracts/scripts/hardhat/eventFilter.ts
@@ -1,0 +1,45 @@
+import { task } from 'hardhat/config';
+import { HardhatRuntimeEnvironment } from 'hardhat/types';
+import { writeFileSync } from 'fs';
+
+task(
+  'grant-role-event-filter',
+  'Filters grant role event for a specific contract to keep track of assigned roles'
+)
+  .addParam('contract', 'address of the contract')
+  .setAction(
+    async (taskArgs: any, hre: HardhatRuntimeEnvironment): Promise<void> => {
+      const provider = new hre.ethers.JsonRpcProvider(process.env.RPC_URL);
+
+      const grantRoleEvent = [
+        'event RoleGranted(bytes32 indexed role, address indexed account, address indexed sender)',
+      ];
+
+      const contract = new hre.ethers.Contract(
+        taskArgs.contract,
+        grantRoleEvent,
+        provider
+      );
+
+      const eventPayload: any = [];
+
+      try {
+        const events = await contract.queryFilter(
+          contract.filters.RoleGranted()
+        );
+
+        events.forEach((event) => {
+          const eventArgs: any = {};
+          eventArgs.role = event.args[0];
+          eventArgs.account = event.args[1];
+          eventArgs.sender = event.args[2];
+
+          eventPayload.push(eventArgs);
+        });
+
+        writeFileSync('grantedRoles.json', JSON.stringify(eventPayload));
+      } catch (error) {
+        throw new Error(`Unable to query events: ${error}`);
+      }
+    }
+  );
--- a/packages/solidity-contracts/scripts/hardhat/eventFilter.ts
+++ b/packages/solidity-contracts/scripts/hardhat/eventFilter.ts
@@ -12,7 +12,74 @@ task(
       const provider = new hre.ethers.JsonRpcProvider(process.env.RPC_URL);
 
       const grantRoleEvenABI = [
-        'event RoleGranted(bytes32 indexed role, address indexed account, address indexed sender)',
+        {
+          inputs: [
+            {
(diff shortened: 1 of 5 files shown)


packages/solidity-contracts/scripts/hardhat/eventFilter.ts
```diff
@@ -0,0 +1,46 @@
...
+        taskArgs.contract,
+        grantRoleEvenABI,
+        provider
+      );
+
+      const eventPayload: any = [];
+
+      try {
+        const events = await contract.queryFilter(
+          contract.filters.RoleGranted()
```
DefiCake, line 28: What happens when you query too many blocks?

DefiCake (commented):
I 'd say that instead of just logging the events, we pick every address that was ever granted a role, and then query hasRole, and filter those that currently have a role assigned. That way we build a summary of currently active roles in the contract.

Ideally we also can match the role IDs against well known roles that we have, like DEFAULT_ADMIN, PAUSER, COMMITTER... and log those role IDs that we do not recognise.

packages/solidity-contracts/scripts/hardhat/eventFilter.ts
viraj124: currently tested with a public rpc, triggering the ci frequently might result in rate limit issues, as currently we query all blocks.

With a private rpc we should be good, but I was also thinking if we can specify a block range too but for now decided to keep the number of inputs minimal, so didn't add that

PraneshASP: imo, hardhat tasks are not the ideal way to do this. Also why do we need this to be in the ci?

viraj124: we need to be able to manually trigger it and there is no tool with a ui live currently that we can use for filtering events, there is a rate limit error risk with the script approach but using a pvt rpc reduces that risk to a great extent

packages/solidity-contracts/scripts/hardhat/eventFilter.ts
```diff
@@ -0,0 +1,118 @@
...
+        );
+
+        for (const event of events) {
```
DefiCake, line 98: I think that you could filter the roles and addresses before hammering the RPC

packages/solidity-contracts/scripts/hardhat/eventFilter.ts
DefiCake: > imo, hardhat tasks are not the ideal way to do this. Also why do we need this to be in the ci?

Please propose alternatives. I am tempted to use external infra for this task, like envio.

PraneshASP: I guess we can use the graph protocol or quickly spin up our own indexer using ponder.sh

packages/solidity-contracts/scripts/hardhat/eventFilter.ts
```diff
@@ -0,0 +1,137 @@
...
+            eventArgs.role =
+              roles.find((role) => role.value === event.args[0])?.name ||
+              'UNKNOWN_ROLE';
+            eventArgs.account = event.args[1];
+
+            eventPayload.push(eventArgs);
+          }
+        }
+
+        writeFileSync('grantedRoles.json', JSON.stringify(eventPayload));
```
DefiCake, line 132:
```suggestion
        writeFileSync(
          'grantedRoles.json',
          JSON.stringify(eventPayload, undefined, 2)
        );
```

Prettify the json

packages/solidity-contracts/scripts/hardhat/eventFilter.ts
```diff
@@ -0,0 +1,137 @@
...
+task(
+  'grant-role-event-filter',
+  'Filters grant role event for a specific contract to keep track of assigned roles'
+)
+  .addParam('contract', 'address of the contract')
+  .setAction(
+    async (taskArgs: any, hre: HardhatRuntimeEnvironment): Promise<void> => {
+      const provider = new hre.ethers.JsonRpcProvider(process.env.RPC_URL);
+
+      const grantRoleEvenABI = [
```
DefiCake, line 14: Maybe use hre.ethers.getContractAt to get access to the typechain types

DefiCake (commented):
The types are a bit all over the place, opened this in my IDE and saw a lot of red

packages/solidity-contracts/scripts/hardhat/eventFilter.ts
viraj124: yeah I looked into it but still got typing errors, seems like a typing bug, will use a fix similar to the one in `upgradeVerification` script

DefiCake (approved):

PraneshASP (approved):

PR #341, opened 2024-10-18
feat: add 6 decimals erc20 token

arthurgeron:
Recently we had a bug in Fuel Explorer's Bridge that was specific to 6 decimals ERC20 tokens. This adds the first  6 decimals ERC20 test token to the ERC20 deployer so we can use it in our E2E and catch such bugs early before production.

### DIFF
--- a/packages/solidity-contracts/test/erc20Gateway.ts
+++ b/packages/solidity-contracts/test/erc20Gateway.ts
@@ -1,6 +1,7 @@
 import chai from 'chai';
 import type { BigNumberish } from 'ethers';
 import {
+  ZeroAddress,
   ZeroHash,
   keccak256,
   parseEther,
@@ -12,6 +13,7 @@ import hre from 'hardhat';
 import type { HarnessObject } from '../protocol/harness';
 import { randomAddress, randomBytes32 } from '../protocol/utils';
 import type {
+  CRY,
   FuelERC20Gateway,
   MockFuelMessagePortal,
   Token,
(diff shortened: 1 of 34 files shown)


packages/solidity-contracts/test/erc20Gateway.ts
```diff
@@ -30,7 +32,11 @@ type Fixture = Pick<
   | 'signers'
   | 'deployer'
   | 'initialTokenAmount'
-> & { fuelMessagePortalMock: MockFuelMessagePortal };
+> & {
+  fuelMessagePortalMock: MockFuelMessagePortal;
+  CRY: CRY;
```
viraj124, line 37: hey, we do have an integration test using 6 decimal tokens (usdt, usdc), here https://github.com/FuelLabs/fuel-bridge/blob/main/packages/integration-tests/tests/bridge_mainnet_tokens.ts

arthurgeron: Looks good but do you mean the test shouldn't be done here or that the token isn't required?
We need 6 decimals assets to be exported as listed assets by local fuel core to test on Fuel Explorer Bridge. Curren't ERC20 Deployer only deploys 18 decimal tokens (ETH and TKN), hence why a third was created.

viraj124: Yeah we don’t need the integration test we already have it

viraj124: @arthurgeron also we have a [script]( https://github.com/FuelLabs/fuel-bridge/blob/main/packages/solidity-contracts/deploy/hardhat/008.set_canonical_token_bytecode.ts) for using the 6 decimals tokens, that we utilize in the current integration test suite so once this script is ran, you'd essentially can utilize the usdc, usdt tokens using their [mainnet addresses](https://github.com/FuelLabs/fuel-bridge/blob/main/packages/solidity-contracts/protocol/constants.ts#L16)

arthurgeron:
Closing since we already have a script that deploys custom assets. Thanks @viraj124

PR #344, opened 2024-10-20
feat: also build docker images for arm

arthurgeron:


arthurgeron:
Closed in favor of #348

PR #353, opened 2024-10-23
fix: grant role event log ci

viraj124:
- adds absolute path in ci for uploading the event logs json file

### DIFF
--- a/.changeset/dirty-hornets-watch.md
+++ b/.changeset/dirty-hornets-watch.md
@@ -1,5 +1,5 @@
 ---
-'@fuel-bridge/solidity-contracts': minor
+'@fuel-bridge/solidity-contracts': patch
 ---
 
 fix grant role event log generation ci
--- a/.changeset/dirty-hornets-watch.md
+++ b/.changeset/dirty-hornets-watch.md
@@ -2,4 +2,4 @@
 '@fuel-bridge/solidity-contracts': minor
 ---
 
-add absolute path for the event logs in ci
+fix grant role event log generation ci
--- a/.changeset/dirty-hornets-watch.md
+++ b/.changeset/dirty-hornets-watch.md
@@ -1,0 +1,5 @@
+---
+'@fuel-bridge/solidity-contracts': minor
+---
+
+add absolute path for the event logs in ci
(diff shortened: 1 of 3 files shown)


packages/solidity-contracts/scripts/hardhat/eventFilter.ts
```diff
@@ -1,6 +1,7 @@
 import { task } from 'hardhat/config';
 import { HardhatRuntimeEnvironment } from 'hardhat/types';
 import { writeFileSync } from 'fs';
+import * as path from 'path';
```
DefiCake, line 4: Is this needed?

.changeset/dirty-hornets-watch.md
```diff
@@ -0,0 +1,5 @@
+---
+'@fuel-bridge/solidity-contracts': minor
+---
+
+add absolute path for the event logs in ci
```
DefiCake, line 5: nit, the path does not look "relative" to me in the CI

packages/solidity-contracts/scripts/hardhat/eventFilter.ts
viraj124: removed was testing something

.changeset/dirty-hornets-watch.md
DefiCake: I meant absolute in my previous comment.

`path: ./packages/solidity-contracts/grantedRoles.json` is a relative path. Absolute paths look like `/path/to/file`,  in other words, they do not start with a `.`

What you did is fix the path as it was relative to the root of the project, whereas before it was being treated as relative to the solidity-contracts directory. The path is still relative :)

DefiCake (dismissed):

.changeset/dirty-hornets-watch.md
```diff
@@ -0,0 +1,5 @@
+---
+'@fuel-bridge/solidity-contracts': minor
```
DefiCake, line 2:
```suggestion
'@fuel-bridge/solidity-contracts': patch
```

PraneshASP (approved):

PR #356, opened 2024-10-24
fix: manual verify upgrade ci

viraj124:
- update manual verify upgrade payload file path in ci

### DIFF
--- a/.changeset/violet-rice-explain.md
+++ b/.changeset/violet-rice-explain.md
@@ -1,0 +1,5 @@
+---
+'@fuel-bridge/solidity-contracts': minor
+---
+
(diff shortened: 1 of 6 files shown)


.changeset/violet-rice-explain.md
```diff
@@ -0,0 +1,5 @@
+---
+'@fuel-bridge/solidity-contracts': minor
```
DefiCake, line 2:
```suggestion
'@fuel-bridge/solidity-contracts': patch
```

PraneshASP (approved):

DefiCake (approved):

PR #357, opened 2024-10-24
chore: update natspec doc for portal reset rate limit method

viraj124:


packages/solidity-contracts/contracts/fuelchain/FuelMessagePortal/v3/FuelMessagePortalV3.sol
```diff
@@ -86,8 +86,8 @@ contract FuelMessagePortalV3 is FuelMessagePortalV2 {
     /**
      * @notice Resets the rate limit amount.
      * @param _amount The amount to reset the limit to.
-     * Fuel's implementation is inspired by the Linea Bridge dessign (https://github.com/Consensys/linea-contracts/blob/main/contracts/messageService/lib/RateLimiter.sol)
-     * Only point of difference from the linea implementation is that when currentPeriodEnd >= block.timestamp then if the new rate limit amount is less than the currentPeriodAmount, then currentPeriodAmount is not updated this makes sure that if rate limit is first reduced & then increased within the rate limit duration then any extra amount can't be withdrawn
+     * Fuel's implementation is inspired by the Linea Bridge design (https://github.com/Consensys/linea-contracts/blob/main/contracts/messageService/lib/RateLimiter.sol)
+     * Only point of difference from the linea implementation is that when currentPeriodEnd >= block.timestamp, then currentPeriodAmount is not updated this makes sure that if rate limit is first reduced & then increased within the rate limit duration then any extra amount can't be withdrawn.
```
DefiCake, line 90:
```suggestion
     * Only point of difference from the linea implementation is that when currentPeriodEnd >= block.timestamp, then currentPeriodAmount is not updated. This makes sure that if rate limit is first reduced & then increased within the rate limit duration then any extra amount can't be withdrawn.
```

DefiCake (approved):
Punctuation

PraneshASP (approved):

DefiCake (approved):

PR #381, opened 2025-01-03
feat: add CI for integration test suite for contract upgrades 

### LINKED ISSUE #206: Implement workflow to run a test suite on a fork for contract upgrades

viraj124:
- closes #206 
- adds upgrade scripts and integration tests that run on a forked L1 network
- removes unused `solidity-contracts`  workspace dependency  in `fungible-token` and `message-predicates` packages to resolve cyclic dependency


packages/solidity-contracts/deploy/hardhat/006._erc721_gateway_v3.ts
```diff
@@ -9,13 +11,33 @@ const func: DeployFunction = async function (hre: HardhatRuntimeEnvironment) {
...
   } = hre;
+
+  const isForking = hre.config.networks[hre.network.name]?.forking?.enabled;
+
   const [deployer] = await ethers.getSigners();
 
-  const fuelMessagePortal = await get('FuelMessagePortal');
+  let portlAddress;
+  if (isForking) {
+    const deploymentPath = path.join(
```
DefiCake, line 21: Can you explain this one? Why do you need to manually file-read the deployments file?

packages/solidity-contracts/deploy/hardhat/009.chain_state_upgrade.ts
```diff
@@ -0,0 +1,101 @@
+import fs from 'fs';
```
DefiCake, line 1: Why do we need to go through an upgrade for a local hardhat deployment?

packages/solidity-contracts/docker/docker-compose.yml
```diff
@@ -0,0 +1,97 @@
+version: '3.4'
```
DefiCake, line 1: I am a bit confused here, what is this orchestration for and why can't it be located in the original docker folder?

DefiCake (commented):
I am not sure I get everything, I might need some more time to review. For now some general questions I have:
- Why do we need to touch the `deploy/hardhat` scripts and perform upgrades in those? In theory all contracts should be already deployed in the fork we are running the tests in.
- There is a lot of duplicated code, for example in the docker setups, it seems like there is a duplication of integration tests orchestration inside the solidity package now.

viraj124:
> I am not sure I get everything, I might need some more time to review. For now some general questions I have:
> 
> * Why do we need to touch the `deploy/hardhat` scripts and perform upgrades in those? In theory all contracts should be already deployed in the fork we are running the tests in.
> * There is a lot of duplicated code, for example in the docker setups, it seems like there is a duplication of integration tests orchestration inside the solidity package now. Why is there a need for that? The way I see it we would need to doubly maintain this code now

- Hey, so instead of performing upgrades in the test, I thought it would be better to have the upgrade scripts which would get triggered only in the forked environment and upgrade the current mainnet contracts, before executing the tests.

- I mentioned the reason for using a separate docker setup in the last comment, I am now looking into solving that so we can use the original docker setup for both forked and local environments

- regarding the integration tests, I think I mentioned this that with the forked environment there were syncing issues related to storage of commit hashes in `FuelChainState` contract, so to avoid more convoluted code in the original integration tests, I decided to have a separate version for the forked environment

does that make sense ?

packages/solidity-contracts/docker/docker-compose.yml
viraj124: hey, this orchestration is for the forked test environment, the idea is to integrate the `testcontainer` package for orchestration in another pr

my original implementation was to use the original docker setup, but post upgrading the `fuel-core` and `block-committer`  versions and having a discussion with the client team.

I realised that for the forked environment, I would need to pass the `--relayer-da-deploy-height` arg when starting the `fuel-core` service so as to make the L2 blocks produce faster, and in the non-forked environment if I ran the original integration tests with `--relayer-da-deploy-height` enabled, the tests got timed out due to not receiving the deposit message when bridging from ethereum to fuel, so that is why I thought it is better to have separate setup for the forked environment.

let me look into adding a env var in the `fuel-core` docker setup, to set the `--relayer-da-deploy-height` only if the fork environment is enable, so then we can just use the original docker setup

packages/solidity-contracts/deploy/hardhat/009.chain_state_upgrade.ts
viraj124: so basically I have made [forking enabled](https://github.com/FuelLabs/fuel-bridge/blob/viraj124/206-ci-fork-integration-test/packages/solidity-contracts/hardhat.config.ts#L44) in the hardhat network, so upgrades are only triggered when forking is enabled otherwise these scripts are not executed.

packages/solidity-contracts/deploy/hardhat/005.register_block_committer.ts
```diff
@@ -7,16 +7,21 @@ const COMMITTER_ADDRESS = '0x9965507D1a55bcC2695C58ba16FB37d819B0A4dc';
 
 const func: DeployFunction = async function (hre: HardhatRuntimeEnvironment) {
   const { ethers, deployments } = hre;
+
+  const isForking = hre.config.networks[hre.network.name]?.forking?.enabled;
+
   const [deployer] = await ethers.getSigners();
 
-  const { address } = await deployments.get('FuelChainState');
+  if (!isForking) {
```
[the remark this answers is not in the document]
viraj124: I'll move the log statement inside the script and make it similar to the previous script, I was doing some testing with this script so that is why it is different but will change to the original pattern.

docker/l1-chain/package.json
```diff
@@ -0,0 +1,85 @@
+{
```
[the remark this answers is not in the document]
viraj124: this is due to the [workspace dependencies](https://github.com/FuelLabs/fuel-bridge/blob/viraj124/206-ci-fork-integration-test/packages/solidity-contracts/package.json#L36-L37).

 I have to use in the `solidity-contract` package, so running `pnpm install` in the L1 docker env throws `ERR_PNPM_WORKSPACE_PKG_NOT_FOUND` error so had to use a version without the workspace dependencies
 
 does that make sense ?

DefiCake:
> * Hey, so instead of performing upgrades in the test, I thought it would be better to have the upgrade scripts which would get triggered only in the forked environment and upgrade the current mainnet contracts, before executing the tests.

I suggest to put these scripts in a separate `deploy/` folder and use that folder when forking. That will make the deploy scripts easier to read and follow. The delta of changes will be reduced too.

docker/l1-chain/package.json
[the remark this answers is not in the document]
viraj124: yeah so initially when I started testing I was a lot of hardhat tools in the test, so that was one of the reasons and then I faced some issues in the tests regarding some storage updates, so I removed the use of hardhat but didn't relocate the tests, so now gonna do that and add the tests under a different folder in the `integration-test` package

viraj124:
> > * Hey, so instead of performing upgrades in the test, I thought it would be better to have the upgrade scripts which would get triggered only in the forked environment and upgrade the current mainnet contracts, before executing the tests.
> 
> I suggest to put these scripts in a separate `deploy/` folder and use that folder when forking. That will make the deploy scripts easier to read and follow. The delta of changes will be reduced too.

will do that

docker/l1-chain/package.json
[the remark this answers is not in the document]
viraj124: reloacted it under the `integration-test` package

packages/test-utils/src/utils/setup.ts
```diff
@@ -109,6 +111,8 @@ export async function setupEnvironment(
     opts.pk_eth_signer1 || process.env.PK_ETH_SIGNER1 || def_pk_eth_signer1;
   const pk_eth_signer2: string =
     opts.pk_eth_signer2 || process.env.PK_ETH_SIGNER2 || def_pk_eth_signer2;
+  const pk_eth_signer3: string =
```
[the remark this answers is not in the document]
viraj124: good catch I was using it in early stages of testing, will remove

packages/integration-tests/fork-tests/bridge_erc20.ts
```diff
@@ -0,0 +1,699 @@
...
+    })
+      .then((res: any) => res.json())
+      .then(async (res) => {
+        if (!res.data.block) {
+          throw new Error(`Could not fetch block with height ${height}`);
+        }
+
+        return res.data.block;
+      });
+  }
```
DefiCake, line 87: This is typeless, why not use `fuelsProvider.getBlock(height)`?

viraj124: replaced

DefiCake (approved):

docker/fuel-core/Dockerfile
```diff
@@ -7,7 +7,8 @@ FROM ghcr.io/fuellabs/fuel-core:v0.40.0
 
 ARG FUEL_IP=0.0.0.0
 ARG FUEL_PORT=4001
-ARG CONSENSUS_KEY_SECRET=""
+# since we need to set the FORKING var in the upgrade-test-suite ci so setting it here
+ARG CONSENSUS_KEY_SECRET="0xa449b1ffee0e2205fa924c6740cc48b3b473aa28587df6dab12abc245d1f5298"
```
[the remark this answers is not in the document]
viraj124: yeah it is a hardhat test account key we already expose it [here](https://github.com/FuelLabs/fuel-bridge/blob/main/docker/envs/fuel_core.env#L2)

PR #385, opened 2025-01-21
test: token withdrawal tests using message coin

### LINKED ISSUE #274: (test) Add a integration test case: withdraw with MessageCoin

viraj124:
- close #274


package.json
```diff
@@ -42,7 +42,7 @@
     "turbo": "^1.10.7"
   },
   "devDependencies": {
-    "@changesets/cli": "^2.26.2",
+    "@changesets/cli": "^2.27.11",
```
DefiCake, line 45: Can you please explain the need for this?

viraj124: the `changeset check` was failing both locally and on the [ci level,](https://github.com/FuelLabs/fuel-bridge/actions/runs/12881356108/job/35912223217) so had to do some debugging and upgrade the package version to fix it

packages/integration-tests/tests/bridge_erc20.ts
```diff
@@ -59,12 +61,88 @@ describe('Bridging ERC20 tokens', async function () {
...
+      transactionRequest.gasLimit = cost.gasUsed;
+      transactionRequest.maxFee = cost.maxFee;
+
+      await fuelTokenSender.fund(transactionRequest, cost);
+
+      // verify that the message coin is consumed
+      const incomingMessagesonFuel = await env.fuel.signers[0].getMessages();
+      expect(incomingMessagesonFuel.messages.length === 0);
+
+      return transactionRequest;
```
DefiCake, line 118: This is 100% using a `InputType.Coin` to fund the message

viraj124: I actually make sure it is the [`MessageCoin` type in the test](https://github.com/FuelLabs/fuel-bridge/blob/viraj124/274-message-coin-withdrawal-test/packages/integration-tests/tests/bridge_erc20.ts#L101), since `Resource` is of `Coin | MessageCoin` type, and `Coin` type does not have [the `daHeight` arg](https://github.com/FuelLabs/fuels-ts/blob/79ffb82d6db638ce50f8bdfdd2a8f8885d12e186/packages/account/src/providers/coin.ts#L8),

DefiCake: It has to be a MessageCoin without other InputType.Coin

viraj124: yeah that's why I added the `daHeight` check, will look into using something alternate than the `Resource` type

viraj124: pushed the update

packages/integration-tests/tests/bridge_mainnet_tokens.ts
```diff
@@ -73,13 +76,107 @@ describe('Bridge mainnet tokens', function () {
   // override the default test timeout from 2000ms
   this.timeout(DEFAULT_TIMEOUT_MS);
 
+  async function fetchTransaction(
```
DefiCake, line 79: This function is repeated three times, I think it should be refactored into an utility.

The name also differs a lot with what it is doing

packages/integration-tests/tests/bridge_erc20.ts
```diff
@@ -59,12 +62,103 @@ describe('Bridging ERC20 tokens', async function () {
...
+          },
+        });
+
+      // verify the incoming messages generated when base asset is minted on fuel
+      let incomingMessagesonFuel = await env.fuel.signers[0].getMessages();
+
+      expect(incomingMessagesonFuel.messages.length === 1);
+      expect(
+        incomingMessagesonFuel.messages[0].amount === fuels_parseEther('1')
+      );
```
DefiCake, line 96:
```suggestion
      expect(incomingMessagesonFuel.messages.length >= 1);
```

Is there a reason why you need that the amount and length are exactly one?

DefiCake (commented):
lgtm, I think it could use some polish

packages/integration-tests/tests/bridge_erc20.ts
viraj124: I bridge 1 eth at the start of the test, so that was the only reason to have a strict equality

packages/integration-tests/fork-tests/bridge_erc20.ts
```diff
@@ -56,12 +59,103 @@ describe('Bridging ERC20 tokens', async function () {
...
+            assetId: fuel_testAssetId,
+          },
+        });
+
+      // verify the incoming messages generated when base asset is minted on fuel
+      let incomingMessagesonFuel = await env.fuel.signers[0].getMessages();
+
+      expect(incomingMessagesonFuel.messages.length === 1);
+      expect(
+        incomingMessagesonFuel.messages[0].amount === fuels_parseEther('1')
```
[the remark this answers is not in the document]
viraj124: the only reason is that the assertions here are dependent on the code exclusive to the function code and avoid repeating the same code i.e fetching the `incomingMessagesonFuel` in this case

packages/integration-tests/fork-tests/bridge_erc20.ts
```diff
@@ -56,12 +59,103 @@ describe('Bridging ERC20 tokens', async function () {
...
+      // fetch tx cost
+      const cost = await fuelTokenSender.getTransactionCost(transactionRequest);
+
+      // update fee params
+      transactionRequest.gasLimit = cost.gasUsed;
+      transactionRequest.maxFee = cost.maxFee;
+
+      // verify that the message coin is consumed
+      incomingMessagesonFuel = await fuelTokenSender.getMessages();
+      expect(incomingMessagesonFuel.messages.length === 0);
```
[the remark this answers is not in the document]
viraj124: replied above

packages/integration-tests/fork-tests/bridge_erc20.ts
```diff
@@ -56,12 +59,103 @@ describe('Bridging ERC20 tokens', async function () {
...
+    fuelTokenSender: FuelWallet,
+    to: string,
+    NUM_TOKENS: bigint,
+    DECIMAL_DIFF: bigint,
+    useMessageCoin: boolean
+  ): Promise<ScriptTransactionRequest> {
+    if (useMessageCoin) {
+      const tx = await fuel_bridge.functions
+        .withdraw(to)
+        .addContracts([fuel_bridge, fuel_bridgeImpl])
```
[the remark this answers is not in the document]
viraj124: yes we do otherwise `ContractNotInInputs` error is thrown, we have had this piece of since the start

DefiCake (approved):

PraneshASP (approved):

